PR #357, opened 2015-01-06
Change MediaPassion URL

tof06:
Media Passion scraper has changed its API URL. There's a 301 redirect on old url, but QNetworkAccessManager doesn't follow it.

Thanks.


Komet:
Thanks for this!

PR #360, opened 2016-01-01
do not require zen and mediainfo for building on mac

boneskull:
These are not necessary; I can build without them.


Komet:
unix: also applies for OSX (it is actually a Unix based OS).
But you're right: -lzen and -lmediainfo are not really needed, it is safe to remove them completely.

boneskull:
@Komet I've rebased with this change in place

PR #361, opened 2016-07-17
Let save functions to mkdir in path

dvdios:
Right now MediaElch can already read and save data in a subdirectory (eg. using "metadata/poster.jpg") but only if the path exists before saving.
Theese changes remove this limitation.


Komet:
Thank you for this addition

PR #364, opened 2017-02-21
Scraper fixes

gnoling:
- hd-trailers changed their markup slightly and now requires an extra space in the regexp that parses out trailer results
- musicbrainz now requires a user-agent for all requests (per their throttling guidelines), which was breaking results for the universal music scraper
- musicbrainz also redirects to the non www url for all requests, so rewrote those to skip the unnecessary redirects


Komet:
Thanks!

PR #373, opened 2018-02-25
Fix some cppcheck errors; use nullptr

archer96:
This PR:

 - uses `nullptr` instead of `0` in many places
 - fixes various cppcheck warnings

_Note_: There are still cppcheck warnings that need to be fixed. Many of them are "unused variable" warnings.


archer96:
This PR breaks MacOS builds (because of a removed `int = 0;`. I will provide a fix soon.

PR #376, opened 2018-03-04
clang-tidy: code cleanup

archer96:
Using clang-tidy, I modernized the codebase in many places. This includes:

 - C++11 `auto`
 - C++11 `override`
 - C++11 `default`
 - C++11 `nullptr`
 - C++11 range for loop
 - C++11 raw string literals
 - Using `empty()` instead of `size() == 0` check
 - a MacOS fix (fixes a bug that was introduced by #373 )

This PR was built and tested on:
 - Linux Mint 18.3 with Qt 5.10
 - [MacBook Pro 2010] macOS  High Sierra with Qt 5.9

I haven't tested this on Windows, yet. I can't figure out how to use/include `zlib` correctly.


archer96:
Apparently some quazip files got "modernized" by accident. I won't revert this for now as quazip needs an update.

PR #389, opened 2018-03-16
Require Qt >= 5.5

archer96:
This PR makes this project require Qt version 5.5 and above in the `.pro` file.
It also removes some checks (Qt >= 5.1.0). Not all checks are removed to avoid merge conflicts with other pull requests.


archer96:
Just found out about `minQtVersion`. Should use that instead.

PR #391, opened 2018-03-20
Escape HTML entities in exported strings

rthouvenin:
Since the data is exported in a HTML file, all the strings that may contain HTML entities (`"`, `<`, `>`, `&`) should be escaped so that they don't break the syntax of the HTML file.

### DIFF
--- a/export/ExportDialog.cpp
+++ b/export/ExportDialog.cpp
@@ -200,20 +200,20 @@ void ExportDialog::replaceVars(QString &m, Movie *movie, QDir dir, bool subDir)
     m.replace("{{ MOVIE.LINK }}", QString("movies/%1.html").arg(movie->movieId()));
     m.replace("{{ MOVIE.IMDB_ID }}", movie->id());
     m.replace("{{ MOVIE.TMDB_ID }}", movie->tmdbId());
-    m.replace("{{ MOVIE.TITLE }}", movie->name());
+    m.replace("{{ MOVIE.TITLE }}", movie->name().toHtmlEscaped());
     m.replace("{{ MOVIE.YEAR }}", movie->released().isValid() ? movie->released().toString("yyyy") : "");
-    m.replace("{{ MOVIE.ORIGINAL_TITLE }}", movie->originalName());
-    m.replace("{{ MOVIE.PLOT }}", movie->overview().replace("\n", "<br />"));
-    m.replace("{{ MOVIE.PLOT_SIMPLE }}", movie->outline().replace("\n", "<br />"));
-    m.replace("{{ MOVIE.SET }}", movie->set());
-    m.replace("{{ MOVIE.TAGLINE }}", movie->tagline());
-    m.replace("{{ MOVIE.GENRES }}", movie->genres().join(", "));
-    m.replace("{{ MOVIE.COUNTRIES }}", movie->countries().join(", "));
-    m.replace("{{ MOVIE.STUDIOS }}", movie->studios().join(", "));
-    m.replace("{{ MOVIE.TAGS }}", movie->tags().join(", "));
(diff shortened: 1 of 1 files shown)


archer96:
:+1: 

Would it be better to escape or remove HTML entities right when setting the movie/tv show properties? So that title, tags, etc. can't contain those characters to begin with?

What are your thoughts on this? 
 I'd say that this step is definitely necessary. I think that `originalName` and other properties should be escaped as well. :smile:

rthouvenin:
In my opinion, it's better to escape than to remove. I submitted this PR because I did encounter some strings with HTML entities, and there were always valid cases. For example, a plot that starts with `On behalf of "oppressed bugs everywhere" an inventive ant named Flik hires...`, or the name of an actor is `Juan "Doc" Restrepo`. 

You are right that other properties should be escaped as well. I thought a genre or tag coming from a movie database would never contain a special character, but I forgot that the data could also be set by the user. Someone may want to have a genre `Romantic <3`, and I think there is no reason to prevent it.
So I think the characters should be kept and escaped.

For `originalName`, it's a mistake, I did want to escape it. Good catch :smile: 

I will update the PR to escape it and other properties, unless you see a reason not to and let me know.

archer96:
Great! :smiley:
I think every property should be escaped. It might not even be the user's fault for having those characters. It's also possible that a website - on which a scraper relies - changes its API and we suddenly have HTML elements in tags, actor names and so on.

I thought of escaping the input string in the setters of the Movie/TvShow classes. But I think this is good for now and might be the better place to escape it. :+1:

(I haven't testet if Qt labels could be affected by HTML entities. I think it would output HTML if the label is surrounded by '<html>')

rthouvenin:
I've updated the PR to escape all QStrings, including in sub-blocks. Let me know if I missed anything.
I squashed the commits for a cleaner history.

Komet:
Looks good, thanks!

export/ExportDialog.cpp
```diff
@@ -200,20 +200,20 @@ void ExportDialog::replaceVars(QString &m, Movie *movie, QDir dir, bool subDir)
...
-    m.replace("{{ MOVIE.TAGS }}", movie->tags().join(", "));
-    m.replace("{{ MOVIE.WRITER }}", movie->writer());
-    m.replace("{{ MOVIE.DIRECTOR }}", movie->director());
-    m.replace("{{ MOVIE.CERTIFICATION }}", movie->certification());
+    m.replace("{{ MOVIE.ORIGINAL_TITLE }}", movie->originalName().toHtmlEscaped());
+    m.replace("{{ MOVIE.PLOT }}", movie->overview().toHtmlEscaped().replace("\n", "<br />"));
+    m.replace("{{ MOVIE.PLOT_SIMPLE }}", movie->outline().toHtmlEscaped().replace("\n", "<br />"));
+    m.replace("{{ MOVIE.SET }}", movie->set().toHtmlEscaped());
+    m.replace("{{ MOVIE.TAGLINE }}", movie->tagline().toHtmlEscaped());
+    m.replace("{{ MOVIE.GENRES }}", movie->genres().join(", ")).toHtmlEscaped();
```
archer96: Should be `m.replace("{{ MOVIE.GENRES }}", movie->genres().join(", ").toHtmlEscaped());` :smiley: 
GCC says: `export/ExportDialog.cpp:210:80: warning: ignoring return value of ‘QString QString::toHtmlEscaped() const’, declared with attribute nodiscard [-Wunused-result]`
Will update this later.

rthouvenin: Good catch, thanks!

archer96: No problem.  :smile: 
It's fixed by #397

PR #394, opened 2018-04-02
Fix OFDb 404 response handling

archer96:
- fix an infinite loop of requests


archer96:
I will resolve these conflicts today.

archer96:
@Komet I fixed all merge conflicts.

PR #396, opened 2018-04-02
Enable gcc warnings; fix warnings

archer96:
- [x] enable (nearly all) warnings for gcc
- [x] fix gcc warnings: c-cast / ...

### DIFF
--- a/MediaElch.pro
+++ b/MediaElch.pro
@@ -1,18 +1,16 @@
-#-------------------------------------------------
-#
-# Project created by QtCreator 2012-02-16T11:13:07
-#
-#-------------------------------------------------
-
 # Check Qt versions
 lessThan(QT_MAJOR_VERSION, 5): error(Qt 4 is not supported!)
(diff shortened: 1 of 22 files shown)


MediaElch.pro
```diff
@@ -29,8 +27,25 @@ win32 {
...
-TEMPLATE = app
+gcc | clang {
+    # Include all Qt modules using isystem so that warnings reagarding Qt files are ignored
+    QMAKE_CXXFLAGS += -isystem "$$[QT_INSTALL_HEADERS]"
+    for (inc, QT) {
+        QMAKE_CXXFLAGS += -isystem \"$$[QT_INSTALL_HEADERS]/Qt$$system("echo $$inc | sed 's/.*/\u&/'")\"
+    }
+}
+
+gcc {
```
archer96: Just found out by using travis-ci, that these warnings are set even if the compiler is clang. I have to fix this...

PR #398, opened 2018-04-05
Add changelog

bugwelle:
I'd like to have a changelog in the root directory (also some sort of beta changelog was requested on http://community.kvibes.de).

The changelog in [MediaElch/debian](https://github.com/Komet/MediaElch/blob/master/debian/changelog) is also not up to date.
Please let me know if this is acceptable or if I should update `MediaElch/debian/changelog`. :smiley: 

This changelog does not contain the full version history. I think for older versions `MediaElch/debian/changelog` is a better place to look.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,0 +1,40 @@
+# Changelog
+
+2.4.3 (*tbd*)
+
+ - Feature: Detect duplicate movies
+ - Feature: Use HTTPS for scraping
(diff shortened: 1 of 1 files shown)


bugwelle:
Updated changelog to list everything up to a7f1123f4a17b798e16090faf7c0e6f59b30b93f

PR #399, opened 2018-04-05
Recognize mk3d as a video file format

bugwelle:
Put simply, it's a 3d mkv file.
See: https://fileinfo.com/extension/mk3d


bugwelle:
Closed in favor of #405

PR #413, opened 2018-05-19
Fix dependencies to build and install on Ubuntu 18.04, while maintain…

danielrheinbay:
…ing compatibility with previous versions:
* build dependency: libcurl4-gnutls-dev as an alternative to libcurl4-openssl-dev
* install dependency: libcurl4 as an alternative to libcurl3

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,14 +2,14 @@ Source: mediaelch
 Section: misc
 Priority: extra
 Maintainer: Daniel Kabel <daniel.kabel@me.com>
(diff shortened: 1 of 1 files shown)


bugwelle:
Thanks :smiley: 

I'm not that familiar with building debian packages, yet. Will this really list `libcurl4` as an alternative to `libcurl3` or will this make both `libcurl3` and `libcurl4` a dependency?

danielrheinbay:
Yes, by using the vertical bar or pipe `|`symbol, you can specify alternative package names as dependencies, of which at least one has to be fulfilled (i.e. boolean OR).
See https://www.debian.org/doc/debian-policy/#syntax-of-relationship-fields for more information.

I was surprised to see these two changes were all that I needed to build MediaElch on Ubuntu 18.04 :)

bugwelle:
Thanks for posting a link to the documentation. :smiley:

PR #467, opened 2018-06-08
Added other extras to be ignored

sharkbit:
Added -behindthescenes, -deleted, -featurette, -interview, -scene, -short to be ignored when scanning for media files.

### DIFF
--- a/data/MovieFileSearcher.cpp
+++ b/data/MovieFileSearcher.cpp
@@ -78,8 +78,10 @@ void MovieFileSearcher::reload(bool force)
 
                 QString dirName = it.fileInfo().dir().dirName();
                 QString fileName = it.fileName();
-                if (fileName.contains("-trailer", Qt::CaseInsensitive)
-                    || fileName.contains("-sample", Qt::CaseInsensitive)) {
+                 if (fileName.contains("-trailer", Qt::CaseInsensitive) || fileName.contains("-sample", Qt::CaseInsensitive) ||
+					fileName.contains("-behindthescenes", Qt::CaseInsensitive) || fileName.contains("-deleted", Qt::CaseInsensitive) ||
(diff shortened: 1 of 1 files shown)


sharkbit:
There might be an easier way to do this like in an array, but a bunch of OR's is pretty easy.

I tend to follow this naming convention:
https://support.plex.tv/articles/200220677-local-media-assets-movies/#toc-3

bugwelle:
For future PRs please run `./scripts/format.sh` before committing changes (if you're on linux or macOS).
The script runs clang-format on the whole repository. Check out https://github.com/Komet/MediaElch/commit/653f9a02980b75e6b1ed691e83bede0c2d65e92e to see how your code was formatted. :smiley: 

We still need to add a batch script for windows, though.

sharkbit:
I use Windows, and was never able to get QT to fully compile the source. So since my changes are usually small I usually just use the web interface in GitHub.   

Thanks, Keep up the great Work!

JoaoCarlos78:
Hi,

Is it possible to change the ignore words for just ignore that word? Ex. as it is it will ignore -scene and -sceneabc, can it be change to just ignore -scene?

Thanks

sharkbit:
Possible, but bond me.  I would suggest changing 'filename.contains' to .endsWith.  Might have to trim off extensions first I dont really know what is included in filename var.

refs:
https://stackoverflow.com/questions/15244911/removing-extension-of-a-file-name-in-qt/15245232
https://het.as.utexas.edu/HET/Software/html/qstring.html#endsWith

bugwelle:
> Hi,
> 
> Is it possible to change the ignore words for just ignore that word? Ex. as it is it will ignore -scene and -sceneabc, can it be change to just ignore -scene?
> 
> Thanks

Sure. What use cases exist for that? :-)
i.e. why don't you use a delimiter such as `-`? :)

PR #470, opened 2018-06-14
[TravisCI] Update to Qt 5.11

bugwelle:


bugwelle:
Qt 5.11 is not available for Trusty using https://launchpad.net/~beineri

PR #477, opened 2018-06-18
Refactor

bugwelle:
I'll try to refactor code in some places. Even if it only means reformatting (e.g. adding braces).
Some classes are way too big and should be split up.

@Komet  I'd like to request your review for this.

---------------

Update: I won't push any more changes to this branch. I used clang-tidy to add braces to all statements, which is enough for now.


bugwelle:
@Komet I've fixed some merge conflicts. Could you please review this PR as my other changes depend on this. :smiley:

PR #483, opened 2018-06-22
Updated Russian translation

alexls74:
Translation of new strings (haven't existed on transifex) and some fixes.


bugwelle:
Thank you for reviewing and updating the translation.
Tansifex doesn't have the latest strings, yet. We still need to update all translations there :smiley:

Issue #485, opened 2018-06-24
Scrolling in settings


alexls74:
The last build has this problem:
![2018-06-24 9 59 25](https://user-images.githubusercontent.com/12618414/41814906-75d4de7c-779e-11e8-82e2-dc48d0619cfb.png)

bugwelle:
I can reproduce this on my Mac using Qt 5.11. Qt 5.10 does not have this problem.
I think that this is a Qt bug.

alexls74:
Changing **verticalScrollMode** from **ScrollPerPixel** to **ScrollPerItem** can fix It.
But I’m not sure that it is good solution.

alexls74:
It seems that the scrolling problem is gone.
```
Compiled with Qt version 5.11.2 (64 Release, Clang 10.0 (Apple))
Using Qt version 5.11.2
System: macOS 10.14 (x86_64-little_endian-lp64)
```
Very nice!

bugwelle:
Oh, nice! 5.11.2 must have fixed it :-)
Will close this issue as soon as I've confirmed this.

PR #498, opened 2018-07-01
[macOS] Fixed Trailer Dialog

alexls74:
This is a really strange solution, but only one I could find to fix this:

Qt 5.11 on macOS is extremely buggy!

### DIFF
--- a/globals/TrailerDialog.cpp
+++ b/globals/TrailerDialog.cpp
@@ -284,6 +284,11 @@ void TrailerDialog::cancelDownload()
 
     m_output.close();
     m_downloadInProgress = false;
+
+#ifdef Q_OS_MAC
+    TrailerDialog::resize(width() + 1,height() + 1);
+    TrailerDialog::resize(width() - 1,height() - 1);
+#endif
 }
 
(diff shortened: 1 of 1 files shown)


alexls74:
Should I close the request?

bugwelle:
I couldn't find a solution for it either. Qt 5.11 is indeed buggy. Should we switch back to 5.10? What do you think.

I'll merge this for now. But we should find a better solution...

alexls74:
> But we should find a better solution...

Sure! This is...  a very awkward solution :(

alexls74:
> Qt 5.11 is indeed buggy. Should we switch back to 5.10? What do you think?

I did It for #485 
But 5.11 on windows seems OK

bugwelle:
Qt works fine on windows. For windows-deployment we're limited to Qt version 5.8 as it is the latest version available on `mxe` (we build for Windows on Linux).
For Linux builds we have to test with Qt version 5.5 because we want to support Ubuntu 14. And on macOS we use the latest version available with `brew`. Qt on mac always looked "different".

PS: It seems that `QVideoWidget` is buggy (there's a lot about it on google). We should look into alternatives (i.e. QML video player).

PR #500, opened 2018-07-06
Media flags and MBAFF scantype

alexls74:
Added video codec mappings and the scantype for MBAFF sources.


settings/AdvancedSettings.cpp
```diff
@@ -31,9 +31,16 @@ void AdvancedSettings::reset()
     m_writeThumbUrlsToNfo = true;
     m_useFirstStudioOnly = false;
 
+    m_videoCodecMappings.insert("v_mpeg4/iso/avc", "h264");
+    m_videoCodecMappings.insert("v_mpegh/iso/hevc", "HEVC");
+    m_videoCodecMappings.insert("27", "h264");
```
bugwelle: This doesn't seem correct... Is there any source on mappings like this other than trial-and-error? :)

settings/AdvancedSettings.cpp
```diff
@@ -272,7 +279,7 @@ void AdvancedSettings::loadFilters(QXmlStreamReader &xml)
  */
 void AdvancedSettings::loadMappings(QXmlStreamReader &xml, QHash<QString, QString> &mappings)
 {
-    mappings.clear();
+    // mappings.clear();
```
bugwelle: Mappings should be cleared! Please add above mappings to `doc/advancedsettings.xml`.

alexls74:
> Why did you delete StreamDetails.cpp in the first place?

Just wrong uploading :)

settings/AdvancedSettings.cpp
alexls74: This is what I got from ML:


Etc.

settings/AdvancedSettings.cpp
alexls74: What is the reason? It’s just clear any preset mappings in AdvancedSettings.cpp
If `<videoCodecs></videoCodecs>` is empty - no mapping at all.

bugwelle: The reason is that someone may not want those mappings. We currently have all default mappings included in `advancedsettings.xml`.

settings/AdvancedSettings.cpp
bugwelle: Thanks. But I think mappings like "27" to "H264" are something the user should add to `advancedsettings.xml` on his own. I just can't find any resource that says that H264 may be encoded as "27" (neither official nor inofficial). I'm also unsure whether this might be a bug by the program used to encode your video files. 
This includes `("2", "MPEG2")`, `("234", "VC-1")` and `("27", "h264")`. 

I'm fine with all other codecs (even though I I couldn't find and source on `microsoft` as well). :smiley:

settings/AdvancedSettings.cpp
alexls74: OK. Deleted.

bugwelle: Thanks :)

PR #505, opened 2018-07-08
[StreamDetails] Codec ID -> Format

alexls74:
"Codec ID" depends on the container format.
For BD sources we can have this:

So... Let’s give up CodecID :)


bugwelle:
Thank you for looking into this. :smiley: 

Side note: I often use "Big Buck Bunny" for testing video codecs, etc.: http://download.blender.org/peach/bigbuckbunny_movies/  
If I run `mediainfo -f big_buck_bunny_720p_h264.mov` I get following output for codec/format:

```
General
Codecs Video                 : AVC
Format                       : MPEG-4
Codec ID                     : qt
Codec                        : MPEG-4
Video
Format                       : AVC
Codec ID                     : avc1
Codec                        : AVC
```
MediaElch uses the Video section for stream details and in this case "Codec ID" is correct. But (at least for this video) there is not Format Version. What I'm trying to say is that we need to check this change for all kind of codecs. We mustn't break anything. :)
I won't have time to test this PR today and it may take a while (so please don't close this PR even if it looks inactive).

I'd also like to request a review from @Komet as I'm not familiar with `mediainfo` and its options, yet.

alexls74:
**Format Version** uses only to define MPEG2

`format = format.toLower();
    if (!format.isEmpty() && format == "mpeg video") {
        format = (version.toLower() == "version 2") ? "mpeg2" : "mpeg";
    }`

bugwelle:
Oh, didn't see that. Thank you.
Okay, I'll test your PR using my movie library as soon as possible. But I'd still like to request a review from Komet, though. :)

alexls74:
Well, back to this question.

> What I'm trying to say is that we need to check this change for all kind of codecs. We mustn't break anything. :)

It's already done by the MediaInfo team :)
https://github.com/MediaArea/MediaInfoLib/tree/master/Source/Resource/Text/DataBase

PR #523, opened 2018-07-14
[Movie] Allow more trailer name variations

### LINKED ISSUE #479: Local Movie Trailers not working
(Windows) the Local Movie Trailer is not showing up as found in the 'Movie List' or the play local trailer now button when the Trailer dose not follow the movie naming convention.

I have a few trailer for each movie and like to name them appropriately  ie RED band, Trailer #3, Trailer #1.

As long as the file ends with -trailer  it should trigger the local trailer.

LOVE THIS APP....

bugwelle:
Fixes #479

### DIFF
--- a/movies/Movie.cpp
+++ b/movies/Movie.cpp
@@ -18,7 +18,6 @@
 Movie::Movie(QStringList files, QObject *parent) :
     QObject(parent),
     m_controller{new MovieController(this)},
-    m_files{files},
     m_rating{0.0},
     m_votes{0},
     m_top250{0},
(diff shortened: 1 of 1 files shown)


sharkbit:
Is the a var to find Just the base Movie Name instead of `arg(fi.completeBaseName());` I have some extra stuff I don't want in the trailer name.  Since it's looking for the Full base name it wants the extras, but plex will show the entire trailer name, so i would like it a bit cleaner.

for example 
Full Movie Name:   Krampus.(2015).1080p.h264.dts.mkv
Trailer:  Krampus.(2015).Trailer.1-trailer.mkv

bugwelle:
I get what you mean. Agreed, having resolution, etc. of the movie in the trailer's filename may confuse users as well. Could you open an issue for this and reference this PR?
I'm currently on vacation and would most certainly forget your comment otherwise. :)

bugwelle:
@sharkbit Didn't forget about your comment :-)
I've opened https://github.com/Komet/MediaElch/issues/551

PR #529, opened 2018-07-18
[Concerts] Fix saving Extra Fanarts

alexls74:


concerts/ConcertController.cpp
```diff
@@ -218,6 +218,9 @@ void ConcertController::onDownloadFinished(DownloadManagerElement elem)
             Manager::instance()->mediaCenterInterface()->imageFileName(m_concert, elem.imageType));
         if (elem.imageType == ImageType::ConcertBackdrop) {
             Helper::instance()->resizeBackdrop(elem.data);
+        } else if (!elem.data.isEmpty() && elem.imageType == ImageType::ConcertExtraFanart) {
```
bugwelle: `!elem.data.isEmpty()` is unnecessary (see line 216).

concerts/ConcertController.cpp
```diff
@@ -218,6 +218,9 @@ void ConcertController::onDownloadFinished(DownloadManagerElement elem)
...
+            Helper::instance()->resizeBackdrop(elem.data);
+            m_concert->addExtraFanart(elem.data);
```
bugwelle: Could you please elaborate why you call `m_concert->addExtraFanart(...);` as well as `setImage()` in line 221?
I had a look at `MovieController.cpp` (line 371) and it seems `setImage` shouldn't be called when you add extra fanart. I would change this to:

```cpp
if (!elem.data.isEmpty()) {
    if (elem.imageType == ImageType::ConcertExtraFanart) {
        Helper::instance()->resizeBackdrop(elem.data);
        m_concert->addExtraFanart(elem.data);
    } else {
        ImageCache::instance()->invalidateImages(
                Manager::instance()->mediaCenterInterface()->imageFileName(m_concert, elem.imageType));
        if (elem.imageType == ImageType::ConcertBackdrop) {
            Helper::instance()->resizeBackdrop(elem.data);
        }
        m_concert->setImage(elem.imageType, elem.data);
    }
}
```

alexls74:
> Was extra fanart just not loaded or not saved on disk?

Downloaded but not saved

> I would change this to: ...

Done

bugwelle:
Thanks! Spotting those bugs is a lot of work. Maybe just a few more months and we're bug free 😄

PR #530, opened 2018-07-19
[Audio Codecs] Add Dolby Digital Plus

alexls74:


bugwelle:
Thanks again! May I ask where you get all those logos from? :-)

alexls74:
They are the official logos  :)
Like already used DD, DTS etc.

bugwelle:
Oh, right. Thanks for clarifying this. :-)

PR #550, opened 2018-08-08
[Windows] TvShows selection background

### LINKED ISSUE #545: [Windows] TV Shows selection
Transparent background for the selected show/season/episode

alexls74:
Fix #545 

This temporary solution is taken from the previous version.

### DIFF
--- a/smallWidgets/TvShowTreeView.cpp
+++ b/smallWidgets/TvShowTreeView.cpp
@@ -225,6 +225,9 @@ void TvShowTreeView::drawRowBackground(QPainter *painter, QStyleOptionViewItem o
         if (alternatingRowColors()) {
             if (index.row() % 2 == 0) {
                 option.features |= QStyleOptionViewItem::Alternate;
+            #ifdef Q_OS_WIN
+                option.features &= ~QStyleOptionViewItem::Alternate;
+            #endif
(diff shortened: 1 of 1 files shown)


smallWidgets/TvShowTreeView.cpp
```diff
@@ -225,6 +225,9 @@ void TvShowTreeView::drawRowBackground(QPainter *painter, QStyleOptionViewItem o
         if (alternatingRowColors()) {
             if (index.row() % 2 == 0) {
                 option.features |= QStyleOptionViewItem::Alternate;
+            #ifdef Q_OS_WIN
+                option.features &= ~QStyleOptionViewItem::Alternate;
```
bugwelle: Seems like you set the style option and unset it again if you're on windows? Wouldn't it make more sense to only call `option.features |= QStyleOptionViewItem::Alternate;` if `#ifndef Q_OS_WIN`? :)

alexls74:
You are right.
`#ifdef Q_OS_WIN
#else
#endif`
Looks better :)
Should I do the new request?

bugwelle:
No, it's okay. :-)
I think a lot of code needs rewriting anyway, so we can do it then.

PR #552, opened 2018-08-18
[Data Model] Split Movie class

bugwelle:
Part 1 of many: Current data models are *huge*. I'll try to split them into smaller ones.


bugwelle:
Introduced a bug with this... Around line 190 in MovieImages.cpp I commented something out which shouldn't be commented out.

PR #595, opened 2018-09-23
[i18n] Update translations

alexls74:
Added new fields and updated Russian translation.


bugwelle:
As always, thanks! :smiley:

PR #635, opened 2018-12-14
[Structure] Create 'data' subdirectory

bugwelle:
This PR does not change any code. It just restructures some folders.

`data` according to https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs#tld.data


bugwelle:
mxe is still down

PR #654, opened 2019-01-22
[macOS] Dark mode disable

alexls74:
Disabling the dark mode in Mojave.


bugwelle:
Thanks! :-)
I'll increase the version number first (change it to `-dev`) and then merge this.

PR #674, opened 2019-02-19
[TvShow] Fix TvShowUpdater crash

### LINKED ISSUE #669: [TvShow] Crash if "Show all episodes" is enabled
Hiho, i loaded 2.60 and start like usual. ME needs starts scanning dirs and most times directly crashes after scan, and sometimes i can click a tv show but then it crashes immediately after i choose "show missing episodes"
Sorry for my bad English ;)
What data do u need to help u finding the bug ?

System: Windows 10

bugwelle:
- MediaElch crashed if "Load missing episodes" was enabled
 - fixes #669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## Next Release (*tbd*)
 The next release will affect your scraper settings. You'll therefore have
-to reconfigure your scraper settings.
+to reconfigure your scraper settings. You may also need to delete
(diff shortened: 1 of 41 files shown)


bugwelle:
There seem to be more issues like this. I'll refactor a bit to avoid string literals like `"tvdb"`.

bugwelle:
The "missing episode" code is quite broken. This may be due to many of my changes. Will take a while.

PR #756, opened 2019-07-27
Enable CONFIG setting for building against external quazip

scottfurry:
Some downstream package maintainers will resort to maintaining patches
against MediaElch sources to enable build against external quazip libraries.

Commit establishes a convenience CONFIG flag:

`CONFIG += USE_EXTERN_QUAZIP`

that enables a DEFINE statement that is used to switch out thirdparty/quazip
references with external quazip sources.

PR #758, opened 2019-07-27
Enable CONFIG setting for building against external quazip(CMake)

scottfurry:
From PR #756, comment was made for need to add commit to CMake Build system files as well.
This commit implements PR#756 into project CMake files.

Executing on the command line:

`cmake ... -DUSE_EXTERN_QUAZIP=ON ...`

will do the same as PR#756, that is internally set a flag in CMake to test, set a define
for project files during compile, and add the library reference to external libquazip5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,11 @@ option(ENABLE_CLANG_TIDY_FIX   "Analyze code with clang-tidy and fix errors."   
 option(ENABLE_COVERAGE         "Add coverage information to binaries."               OFF)
 option(ENABLE_COLOR_OUTPUT     "Force produce ANSI-colored output (GNU/Clang only)."  ON)
 option(ENABLE_LTO              "Enable link-time-optimization. Increases link time." OFF)
+option(USE_EXTERN_QUAZIP       "Enable link-time-optimization. Increases link time." OFF)
+
+if(USE_EXTERN_QUAZIP)
+  add_definitions(-DEXTERN_QUAZIP)
+endif()
 # cmake-format: on
 
 find_package(Sanitizers)
@@ -109,11 +114,16 @@ find_package(
 # Subdirectories and main executable
 
 add_subdirectory(docs)
-add_subdirectory(third_party EXCLUDE_FROM_ALL)
+if(NOT USE_EXTERN_QUAZIP)
+  add_subdirectory(third_party EXCLUDE_FROM_ALL)
+endif()
 add_subdirectory(src)
 
 add_executable(mediaelch src/main.cpp)
 target_link_libraries(mediaelch PRIVATE libmediaelch)
+if(USE_EXTERN_QUAZIP)
+  add_library(quazip5 SHARED IMPORTED)
+endif()
 set_target_properties(mediaelch PROPERTIES OUTPUT_NAME "MediaElch")
 mediaelch_post_target_defaults(mediaelch)
 
--- a/cmake/coverage.cmake
+++ b/cmake/coverage.cmake
@@ -6,6 +6,12 @@ include(${CMAKE_CURRENT_LIST_DIR}/Gcov.cmake)
 
 set(COVERAGE_ENABLED OFF BOOL)
 
+if(NOT USE_EXTERN_QUAZIP)
+  SET(LCOV_THIRDPARTY "\"*/third_party/*\"")
+else()
+  SET(LCOV_THIRDPARTY "\"*/third_party/catch2/*\"")
+endif()
+
 set(
   LCOV_EXCLUDE_COVERAGE
   ${LCOV_EXCLUDE_COVERAGE}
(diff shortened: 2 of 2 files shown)


scottfurry:
Pushed altered coverage setting.
Original PR was 'aggressive' in removing `thirdparty` folder but other files present that are used.

cmake/coverage.cmake
```diff
@@ -6,6 +6,12 @@ include(${CMAKE_CURRENT_LIST_DIR}/Gcov.cmake)
 
 set(COVERAGE_ENABLED OFF BOOL)
 
+if(NOT USE_EXTERN_QUAZIP)
+  SET(LCOV_THIRDPARTY "\"*/third_party/*\"")
```
bugwelle: +1, thanks :)

CMakeLists.txt
```diff
@@ -109,11 +114,16 @@ find_package(
 # Subdirectories and main executable
 
 add_subdirectory(docs)
-add_subdirectory(third_party EXCLUDE_FROM_ALL)
+if(NOT USE_EXTERN_QUAZIP)
+  add_subdirectory(third_party EXCLUDE_FROM_ALL)
```
bugwelle: -1, because there may be third_party code besides quazip. I'd prefer to handle this in `third_party/CMakeLists.txt`. Though, that file only contains QuaZip related stuff at the moment. :)

CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ option(ENABLE_CLANG_TIDY_FIX   "Analyze code with clang-tidy and fix errors."
 option(ENABLE_COVERAGE         "Add coverage information to binaries."               OFF)
 option(ENABLE_COLOR_OUTPUT     "Force produce ANSI-colored output (GNU/Clang only)."  ON)
 option(ENABLE_LTO              "Enable link-time-optimization. Increases link time." OFF)
+option(USE_EXTERN_QUAZIP       "Enable link-time-optimization. Increases link time." OFF)
+
+if(USE_EXTERN_QUAZIP)
+  add_definitions(-DEXTERN_QUAZIP)
```
bugwelle: I try to avoid non-target-specific functions. I would prefer to move this to the `mediaelch_post_target_defaults` function that is applied to each of MediaElch's targets :)

scottfurry: This is that 'modern CMake' thing, isn't it?
Bit of a grey area for me, personally. But after your comment and looking at the CMakeLists.txt, I see your point - the define setting should be in the function(~line 60).
I can rework and resubmit later. My mess to clean this up.

CMakeLists.txt
scottfurry: Yeah, I agree.
I initially had excluded the whole third party folder then came back and changed it because I thought that there might be other folders in there(discovered on catch2). I'll have to look at if there is a `NOT` glob wildcard that could be used.

cmake/coverage.cmake
scottfurry: Tried to do the coverage testing but I kept getting an empty folder for output to genhtml.
I wasn't sure if it was some setting I was missing in doing  the setup, especially after reading through the README (`ninja coverage`).

CMakeLists.txt
bugwelle: Yeah, exactly! "Modern" CMake is what I try to use (we require CMake v3.13 after all)
Don't worry, I'll push a few changes in a few minutes (after I successfully compiled MediaElch with my system's QuaZip library) :)

cmake/coverage.cmake
bugwelle: `ninja coverage` works on my machine but e.g. not on TravisCI, so it may be very likely a bug on our side if it doesn't work for you. :)

CMakeLists.txt
scottfurry: I had to update my CMake to v3.15 to work this(was v3.12). It's kinda of cool reading some of the theory. I get the premise. Implementation and general practice is an entirely different. Good to have a working example as a reference.

cmake/coverage.cmake
scottfurry: This sounds like a missing `something` in the settings to me. Some little nugget set in a user profile that makes everything 'just work' but hasn't been migrated to project configuration. But that's a guess on my part.

CMakeLists.txt
bugwelle: Well, I try to have this as modern as it can be. Though there are a few places where we may not exactly be "best practice". e.g. we use `mediaelch_post_target_defaults` which I think isn't best practice. But it works... :)

cmake/coverage.cmake
bugwelle: Yeah, I'll investigate that issue next week after my vacation (I won't be able to read issues for the next ~1week) :)

CMakeLists.txt
scottfurry: But makes phenomenal sense in hindsight. I can see the utility here.

bugwelle:
@scottfurry Just wanted to inform you that the next MediaElch version will search for QuaZip headers in `quazip/` and no longer `quazip5`. By doing that, we can also run with QuaZip 1.1

See https://github.com/Komet/MediaElch/issues/1339

Arch already uses QuaZip 1.1: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/mediaelch

PR #759, opened 2019-07-28
[CMake] Use modern CMake for external QuaZip usage

bugwelle:
- related #758
 - fix some clang-format issues
 - fix some cmake-format issues

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,10 +33,6 @@ option(ENABLE_COVERAGE         "Add coverage information to binaries."          
 option(ENABLE_COLOR_OUTPUT     "Force produce ANSI-colored output (GNU/Clang only)."  ON)
 option(ENABLE_LTO              "Enable link-time-optimization. Increases link time." OFF)
 option(USE_EXTERN_QUAZIP       "Enable link-time-optimization. Increases link time." OFF)
-
-if(USE_EXTERN_QUAZIP)
-  add_definitions(-DEXTERN_QUAZIP)
-endif()
 # cmake-format: on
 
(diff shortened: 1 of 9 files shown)


bugwelle:
@scottfurry FYI, this compiles on Ubuntu but I have some issues using MediaElch which may not even be related to your and my changes. I may not have time to dig in to that this week, though.

scottfurry:
Compiling, running tests, and executing end product were done on my end (self-QA/QC dept.).
Everything seems to run fine on Funtoo (gentoo - derivative) using Qt 5.12.3. Although, FTR, there are some warnings that pop up during compile above looping and assignments.

I have cppcheck and I have a marginal idea of how to use it. ;)

bugwelle:
Hi, just tested older commits before your changes and it seems to be another issue not related to this PR.

> some warnings that pop up during compile above looping and assignments

Yeah, those pop up in release builds. I think most of them are false-positives as the C++11 for each loop should work with Qt containers, so I have no idea why those warnings pop up. :)
I'll most likely just disable them. :)

scottfurry:
@bugwelle - FYI - https://www.embeddeduse.com/2016/05/13/simplifying-loops-with-cpp11/
Discussion of range base for loops and some ideas for usage w/ QT containers.

bugwelle:
Thanks.

> When we move to C++11 in our projects, we should consistently use range-based for instead of index-based loops, iterator-based loops or foreach macros. Based on my experience, we should be able to do this for 95% of the for-loops.

I have replaced most foreach macros with range based for loops. :)
I will look at those warnings again soon. :)

PR #765, opened 2019-08-08
Correct description text in CMake file

scottfurry:
Issue #756 was incomplete. Description was copy/pasted from above line
but not updated to reflect intent. Edited text.


bugwelle:
Wow. My reviews are useless it seems :smile: 
Thanks! :)

scottfurry:
I'm sitting here shaking my head that I pushed it in the first place. You're commit heads up had the line in question. It reached out from the monitor and slapped me in the face shouting "What were you thinking?". I had to fix...

bugwelle:
Oh, I know that feeling too well :-)
I'll merge this as soon as TravisCI finishes.

PR #780, opened 2019-09-05
[Package] Require Ubuntu 16.04; update to Qt 5.12 for AppImages

bugwelle:


bugwelle:
Tested on my custom branch and bintray account ( https://bintray.com/bugwelle/MediaElch/MediaElch-linux ). Works well. We have to update because linuxdeployqt now only supports Ubuntu 16.04 and later. Ubuntu 14.04 is not supported anymore. I've updated our docs accordingly.

bugwelle:
Yeah, of course.. Updating `install_dependencies.sh` break our Qt 5.5. build.

bugwelle:
Aaaaand there's a typo: `suer` should have been `use`. Well, whatever.

Issue #728, opened 2019-06-15
Loading Streamdetails from file does not work properly


nkatsidis:
Latest Nightly (8/6/2019)

Click a movie -> Streamdetails ->Reload from file

Audio does not show language and type (it shows only number of channerls)
subtitles do not show language

nkatsidis:
also codec now from h264 goes to avc.. is it normal?

nkatsidis:
problem still there in 16/6/2019 version

ticao2:
I believe the answer is here:  
https://github.com/Komet/MediaElch/issues/695

nkatsidis:
i replaced mediainfo.dll 64bit version but the problem is still there...

nkatsidis:
using windows 10 64bit

azriel84:
I have the same issue, windows 10 64bit, I just get 7.1 on sound streamdetails, no Atmos or DTS etc...

nkatsidis:
problem still there in 20/6/2019 nightly

alexls74:
Some fields have changed in mediainfo (>18.05).
We can back to version 18.05 or adapt to new versions.

nkatsidis:
i can confirm that after using 18.03 version of Mediaelch.dll steamdetails seem to work - almost - properly..
BUT... in Codec it does not display h264 but avc..... ;)

azriel84:
Plus latest nightly has issue with DTS:X 7.1 sound stream. Freshly released Hulk (2003) has that stream but mediaelch doesn't depict any sound details at all.

bugwelle:
I'll consider switching back to an older version of MediaInfo until I've figured out what exactly changed.

@alexls74 Do you maybe know if v17 still worked as expected?

bugwelle:
I'll close this issue. In case that streamdetails continue to not work, feel free to reopen it. :)

nkatsidis:
Windows 10 64 bit. Latest Nightly

Loading Streamdetails still not working properly...
I have a Movie with atmos, in audio it does not displays what the audio is (dtshd, truehd, ac3 etc)

version 260 works fine ....

monisriz:
AAC audio codec is not written to the nfo. It's not being shown in UI either.

bugwelle:
For reference: #851 fixes some other audio issues :)

PR #786, opened 2019-09-12
[StreamDetails] Add "avc" to "h264" default codec mapping

bugwelle:
- #728


bugwelle:
The string change needs to be dropped. Doesn't work on Windows..

Issue #781, opened 2019-09-06
[Music] Use correct CD art filename: discart.png


bugwelle:
**Describe the bug**
The correct filename is `discart.png` and not `cdart.png`.

**To Reproduce**
n/a

**Expected behavior**
Use `discart.png`

**Screenshots**
n/a

**MediaElch Version:**
 - [x] 2.6.1-dev (nightly)
 - [x] 2.6.0 stable

**Operating System:**
 - [x] Windows
 - [x] macOS
 - [x] Linux (distribution: ___)

**Additional context**
https://forum.kodi.tv/showthread.php?tid=346924

bugwelle:
https://github.com/Komet/MediaElch/issues/781

bugwelle:
![discart](https://user-images.githubusercontent.com/1667306/64871312-43cb5f80-d645-11e9-98fb-3995383e3a9f.png)

Users can already change the filename :-)

PR #799, opened 2019-10-02
Fix/Rename for Kodi artwork types

monisriz:
This fixes most of the gaps in the Media Elch's naming scheme for the artwork for different media types.
For more info refer to https://kodi.wiki/view/Artwork_types

Note: There are still some artwork types supported by Kodi but not by Media Elch. But most of the common ones should be covered.

Resolves: https://github.com/Komet/MediaElch/issues/781

### DIFF
--- a/src/settings/Settings.cpp
+++ b/src/settings/Settings.cpp
@@ -30,42 +30,42 @@ Settings::Settings(QObject* parent) : QObject(parent), m_advancedSettings{new Ad
     m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieNfo, "<baseFileName>.nfo", 0));
     m_initialDataFilesFrodo.append(DataFile(DataFileType::MoviePoster, "<baseFileName>-poster.jpg", 0));
     m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieBackdrop, "<baseFileName>-fanart.jpg", 0));
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieCdArt, "disc.png", 0));
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieClearArt, "clearart.png", 0));
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieLogo, "logo.png", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieCdArt, "<baseFileName>-discart.png", 0));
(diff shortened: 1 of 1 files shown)


bugwelle:
Awesome! One thing we have to be aware of is whether v16 has a different naming scheme as I still try to support it. Frodo (see variable names) is unsupported, though. I'll merge this tomorrow if v16 has the same naming conventions 👍

monisriz:
As far as I know v16 does not support extended artwork out of the box without the use of an addon (e.g. Artwork Beef). v18 natively recognizes extended artwork by "activating" these artwork types in advancesettings.xml. So the basic artwork should work with v16.

Even if there are any naming differences, users can always manually edit these in the settings pane if they are using a legacy version of Kodi.

bugwelle:
Yeah, we don't yet have a "use v16 names" option or something like that. And you're right that users can still change the file names. :-)

src/settings/Settings.cpp
```diff
@@ -30,42 +30,42 @@ Settings::Settings(QObject* parent) : QObject(parent), m_advancedSettings{new Ad
...
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieLogo, "logo.png", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieCdArt, "<baseFileName>-discart.png", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieClearArt, "<baseFileName>-clearart.png", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieLogo, "<baseFileName>-clearlogo.png", 0));
     m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieBanner, "<baseFileName>-banner.jpg", 0));
     m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieThumb, "<baseFileName>-landscape.jpg", 0));
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieSetPoster, "folder.jpg", 0));
-    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieSetBackdrop, "fanart.jpg", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieSetPoster, "<baseFileName>-poster.jpg", 0));
+    m_initialDataFilesFrodo.append(DataFile(DataFileType::MovieSetBackdrop, "<baseFileName>-fanart.jpg", 0));
```
bugwelle, line 39: I didn't properly review this... This should be `<setName>-...`

PR #803, opened 2019-10-05
[Style] Fix label color not shown for movies

bugwelle:
- Remove `border:none` from QTableWidget
 - Note from Qt docs (QTableWidget):
   https://doc.qt.io/qt-5/stylesheet-reference.html

   "If you only set a background-color on a QTableCornerButton,
   the background may not appear unless you set the border property
   to some value. This is because, by default, the QTableCornerButton
   draws a native border which completely overlaps the
   background-color."
 - bug introduced in 629786b2b4ee92c8ab6495b2ae2ffcdf243c3d64
   (Mon Feb 18 08:45:19 2019 +0100)

### TODO
 - [ ] Test if it still works with a dark theme as we partly revert "[UI] Fix table/list layout for dark theme"


bugwelle:
Oh.. Merged the wrong pr. Will have to check if dark node breaks.

PR #882, opened 2020-02-04
[ImageProvider] Show error if provider fails retrieving images (e.g. network failure)

### LINKED ISSUE #874: Change message in 'Choose an Image' on time out
**Is your feature request related to a problem? Please describe.**
Now you get the same message "No images found" in the 'Choose an Image'-dialogue no matter what the reason is that no images are shown.

**Describe the solution you'd like**
Only show the "No images found" when there really are no images found and display "Source timed out" or something like that when ME fails to get a response. Using fanart.tv at the moment is a lottery so would be nice to know if the site is worth a visit or not to download the images manually.

**Describe alternatives you've considered**
Skip handling fanart in ME and do it manually, not that much of a difference in how much time it takes at the moment since you have to double check it anyway if you get no results after waiting a minute or two.

**Additional context**
Think it's a good idea to move #834 and #835 up a bit on the to do list since fanart.tv seem to be dying. Site maintenance is done by a single moderator, none of the admins have been around for six months, the owner even longer, and their servers appear to be on their last breath.

bugwelle:
Fixes #874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@
  - AdvancedSettings: Add experimental exclude patterns (#840)
  - Add `en_US` language file for better singular/plural handling
  - Better network error reporting for scraping TV shows and movies (#870)
- - Better error reporting in the image dialog (#864)
+ - Better error reporting in the image dialog (#864, #874)
 
 ### Internal Improvements and Changes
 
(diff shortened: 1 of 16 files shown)


bugwelle:
Codecov timed out... Merging.

PR #892, opened 2020-02-10
[Scraper] Add new MovieScraper interface

bugwelle:
## What is this?
The current MovieScraperInterface has multiple issues. This PR will add a new and improved interface. 

### Dependency on UI
The current (movie) scrapers all depend on the UI, i.e. they all add their dropdowns, textfield, etc. to the settings UI. This is an anti-pattern. The business logic should not have anything to do with the UI. A proper solution would be to have a settings class that encapsulate different settings mechanisms (e.g. "select one of many", "textfield", "number", ...).

### Bad error handling
I've tried to improve error handling in the past but the current solution doesn't satisfy me at all. The new interface should not mix success and error signals.

### All In One
All scrapers currently do all-in-one: Downloading + processing. Maybe we should split that up.

### Dependencies on Singeltons
We have quite a few singletons. I don't like that and we should think about how we can get rid of them. We may install a custom logger, etc. 
Or a network tracker (to cache and manage downloads).

### Dependency on QSettings, etc.
The title says it all. Even though we can abstract it (and so so by using `ScraperSettings`) it's more tedious to test it than it should be.

## How the new solution works
I had a look at TMM ( https://gitlab.com/tinyMediaManager/tinyMediaManager/-/tree/devel/src/main/java/org/tinymediamanager/scraper ). I like many of their ideas. But their coupling of TMDb and IMDb is a bit too much IMHO.

TODO

## TODO
- [x] write a proper interface (duplicate existing one)
- [x] transform one scraper
- [x] write tests for that scraper
- [ ] Integrate into command line interface (`mediaelch_cli`)
- [ ] check how it works out (settings, etc)
- [ ] integrate new interface into UI, replace old interfaces
- [ ] Transform all old scrapers to new interface
- [ ] Remove old interface and use new one


bugwelle:
I would also like to refactor most of the movie scrapers, however, that is more work than I can currently handle. Therefore it will mostly just be an interface change.

bugwelle:
The new interface works quite well. Error handling has improved, the code if the TMDb scraper is more readable. Next steps:

- [ ] Integrate into command line interface (`mediaelch_cli`)
- [ ] Remove old interface and use new one
- [ ] Transform all old scrapers to new interface

bugwelle:
What makes this a bit harder is that a movie scrape job looks currently like this:

 - MovieSearch -> Open Dialog
 - MovieSearchWidget -> Load search results using scraper + Show results
 - MovieSearchWidget -> result clicked -> accept MovieSearch Dialog
 - MovieSearchWidget -> tell movie controller to `loadData`
 - MovieController -> use scraper to load data
 - Scraper -> loads information
 - MovieController -> emit loadDone signal
 - MovieWidget -> show results

This is just too nested. I'm fine with the controller loading data using a scraper but we need to make it more structured.

txtsd:
Any update to the On Hold status?
tinymediamanager switched to a subscription model so I'm looking to switch over, but I need to be able to choose TMdb for both movies and shows.

bugwelle:
@txtsd I will have to rework this PR completely. But on the good side, https://github.com/Komet/MediaElch/pull/972 is nearly ready to go. I'm currently writing the last few tests and fix some minor bugs, but all in all it looks promising.
Can't comment on the time line. I didn't have any time for this project in the last 3 months so I'm way behind my schedule. :-/

I'm also surprised that TMM switched to a subscription model. 😮

txtsd:
Oh good. Hopefully I'll be able to understand the code and contribute.

bugwelle:
> Oh good. Hopefully I'll be able to understand the code and contribute.

That would be awesome. If you have any questions, you can also send me an eMail. I've also added some documentation about the new scrapers (what? why? how?) 😃

ticao2:
> 
> 
> > Oh good. Hopefully I'll be able to understand the code and contribute.
> 
> I've also **added some documentation about the new scrapers** (what? why? how?) 😃

Where? I can read?  **:-)**

bugwelle:
Can be found in another PR. I need to update that document, though. A bit has changed. :-)

https://github.com/Komet/MediaElch/blob/bdb047eece6619c445cc3f3dd65349a9fa09ecc4/docs/contributing/scraper-interface.md

bugwelle:
Will update the movie scraper step by step now. Similar to our TV scrapers. This PR can be closed.

PR #912, opened 2020-03-08
[Dev] Set HTTP User-Agent header; Minor refactorings

bugwelle:


bugwelle:
macOS build hangs. Works locally.

PR #949, opened 2020-04-08
[Dialogs] Fix downloading trailers for many movies and tv tunes for shows

### LINKED ISSUE #940: Movie Trailer Downoad Breaks after no trailers foiund on previous movie
When downloading movie trailers, the window will not load properly if the last movie did not return any trailer.

**To Reproduce**
1. Pick a movie that is obscure or might not have trailer (IE Krull 1983)
2. Open Download Trailer.. 
3. Nothing Found in Trailer Download window
4. Close Window
5. Pick a Movie that will have trailer
6. Trailer Download window dose not load fully

**Screenshots**
Step 6 broke window

**MediaElch Version:**
- MediaElch_win_2.6.5-dev_2020-03-08_18-00_git-master-6e27304

**Operating System:**
 - Windows 10 v 1909

bugwelle:
Fix #940

**[TrailerDialog] Fix downloading trailers for many movies**  
If the trailer dialog window was closed using the window frame's close
button ("x") instead of the button labeled "Close", the dialog was
destroyed and reopening it resulted in an empty dialog window.

There is no need to use a singleton for the TrailerDialog so it was

**[TvTunes] Fix crash when dialog is closed and re-opened for other show**  
If the TvTunes dialog is opened the first time for a show, it downloads
some resources. This download does not stop when the dialog is closed.
If the dialog is now reopened (e.g. for another show) we get a race
condition because two threads try to dequeue a list with only one item.
Because an empty list cannot be dequeued, `abort()` is called by Qt.

The fix is simple: Use one TvTunes instance for each show download.  We
can achieve that by simply creating a new TV Tunes dialog for each show.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next Release (*tbd*)
 
+*Note:* This release has changed its internal file and directory handling to
+fix Windows-related bugs (#732).  Please report any issues with network drives
+or other non-local drives.
+
 ### Bugfixes
 
  - Fix AEBN crash when scraping a movie (#910)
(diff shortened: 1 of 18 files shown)


bugwelle:
Build fails with Qt 5.5 due to `QOverload`. Fix will be uploaded soon.

PR #971, opened 2020-05-02
[Movie] Allow screenshot for movie backdrop

### LINKED ISSUE #965: Movie Frame as Fanart - Movies
It would be great if you could add the ability to save a frame of the movie as a fanart. (it is something that is already implemented for Tv Episodes and Posters)

thank you
Nicolas

bugwelle:
Fix #965


bugwelle:
Tested, works.

Issue #847, opened 2019-12-30
[Dev] MediaElch Development Roadmap 2023


bugwelle:
*This is an often-updated issue that exists for discussing decisions regarding MediaElch's development and a continuation of #743.*

## What is this issue?
This issue lists and describes planned features and other design decisions. It is intended for MediaElch's developers and may go into technical details. All of these issues are neither bug reports nor feature requests. They're not big enough to create new issues for them. But other developers may want to try and fix one of the issues below so I thought: Why not share my thoughts?

It's also easier to keep track of one issue with all development design decisions than to have multiple smaller issues.

### Table of Contents
(order: importance from highest to lowest)

 1. Separation of UI and Business Logic
 1. ~~New movie scraper interface~~
 1. ~~New TV show scraper interface~~
 1. ~~Remove singletons for dialogs~~ 
 1. Test file searchers
 1. CMake Windows build
 1. Continuous Deployment
 1. ~~Use CTest and CDash~~
 1. Test full-disk ("no space left") situations
 1. CMake Windows Wix Package
 1. Add macOS Updater
 1. ~~Better logging~~
 2. Better Parallel Filesearch

### Separation of UI and Business Logic
As of 2019-12-30 MediaElch's UI and business logic are still highly coupled. This has quite a few draw backs like:

 - unit testing requires the GUI to load
 - `mediaelch_cli` is dependant on `Qt5Widgets`

As an example the TvShowFileSearcher depends on the TvShowFilesWidget to display all shows. This should be decoupled.

### ~~New movie scraper interface~~
~~See https://github.com/Komet/MediaElch/pull/1163~~

### ~~New TV show scraper interface~~
~~See https://github.com/Komet/MediaElch/pull/1163~~

### ~~Remove singletons for dialogs~~
~~Singletons need to be initialized at startup. This is not needed for dialogs like "TvShowSearch.ui". It only makes MediaElch start slower and makes it use more RAM.~~

See https://github.com/Komet/MediaElch/pull/980

### Test file searchers
We need test cases for testing our file searchers. How could this be done?
We already have a bash script that generates a lot of movie/tv show/concert/artist subfolders that can be used to test MediaElch. We could load all details of every media entry and store all of that in a `.tar.gz` on Dropbox (or whatever cloud hoster there is). The `.tar.gz` can then be downloaded for automated testing.

May be useful: https://kodi.wiki/view/Samples

### Test full-disk ("no space left") situations
[...]

PR #980, opened 2020-05-13
[Search Dialogs] Remove singleton for dialog

bugwelle:
Singletons need to be initialized at startup. This is not needed for dialogs like "TvShowSearch.ui". It only makes MediaElch start slower and makes it use more RAM.

See https://github.com/Komet/MediaElch/issues/847


bugwelle:
Tested locally. Everything seems to work just fine. :)

PR #991, opened 2020-06-06
Update Database.cpp

fialky:
Fix - labels were not deleted from MediaElch.sqlite during the Reload all Movies (F5) after the movie deletion from a folder.

### DIFF
--- a/src/data/Database.cpp
+++ b/src/data/Database.cpp
@@ -289,6 +289,10 @@ void Database::commit()
 void Database::clearAllMovies()
 {
     QSqlQuery query(db());
+    query.prepare("DELETE FROM labels");
+    query.exec();
+    query.prepare("DELETE FROM sqlite_sequence WHERE name='labels'");
+    query.exec();
     query.prepare("DELETE FROM movies");
     query.exec();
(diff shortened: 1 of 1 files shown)


bugwelle:
Hi,

first of all: Thank you for this contribution. It is very much welcome 👍 

I haven't changed anything in the database in the last two years, so my memory isn't the best. But `labels` was responsible for coloring movies, etc.
I don't think we want to have to re-set the color-label for each movie after a reload. I may be mistaken, though. I'll look into it tomorrow.


Regards,
Andre

fialky:
I guess that you know it already: Every "Reload all Movies (F5)" movies, movieFiles, movieSubtitles tables are truncated in MediaElch.sqlite. Therefore I also added labels table for truncate.

bugwelle:
Hi,

just had a look at this PR and tested it. I will not merge it for the following reason: Users can spend quite the amount of time tagging each movie with a color label. If they want to search for a new movie end click "Reload (F5)" then all the manually assigned labels will disappear with your change (I tested and confirmed that).

I don't think that's what most users want. It's not what I want for my library.

The change was a good idea. Our database code needs to be refactored. It is a mess. 😃

PR #1005, opened 2020-08-18
[Docker] Add missing libssl-dev dependency for mxe-cmake

bugwelle:
The update to CMake 3.17 in mxe produced this issue. We were simply missing some headers.


bugwelle:
Ok, so CMake builds just fine but boost fails due to a compiler bug. Will merge it, though.

Issue #1002, opened 2020-07-09
Fix antipattern when loading translations in Qt


bugwelle:
See https://www.kdab.com/fixing-a-common-antipattern-when-loading-translations-in-qt/

We actually use the anti-pattern, though through a hierarchy of functions.

bugwelle:
Fixed by #1018

bugwelle:
Re-open. The current implementation is faulty.

bugwelle:
We do it the correct way, but there is a nasty Qt bug, which was fixed in Qt6: https://github.com/Komet/MediaElch/issues/1479

PR #1018, opened 2020-09-26
[i18n] Use proper translation loading

bugwelle:
The article at [1] describes how to properly load translations.  While
MediaElch did load it partially correct, we had some special coding
that is now removed.

[1]: https://www.kdab.com/fixing-a-common-antipattern-when-loading-translations-in-qt/


bugwelle:
Fixed #1002

PR #1054, opened 2020-11-18
[Scripts] Add package script for macOS

bugwelle:
Used by admins when TravisCI removes free support for macOS.


bugwelle:
This may break the release scripts for Debian/Ubuntu. Will have to check this.

PR #1084, opened 2020-12-15
[CI] Refactor CI scripts to make it possible to build locally as well 

bugwelle:
Because we can no longer rely on TravisCI, I took the time to refactor the packaging scripts.

I want to be able to simply run a few commands from my MacBook and release a Nightly on Bintray. That should work because we can use Docker for Windows and Linux builds.

TODO:

- [x] Windows: unzip may need to overwrite files. Should not abort
- [x] Windows: New ffmpeg hash `d7ced7d203dc689c53bbbf6293e5f68ecafda33e46057fc386e85e6d77c00ac52926df39adaf5235f031cf6199f11603a81b0a8d2e3946957be4dd7e5750ae5c` and version `ffmpeg-4.3.1-2021-01-01-essentials_build`
- [x] MediaElch.app folder should be deleted before building for macOS or macdeployqt will fail
- [x] macOS: Move .dmg from build to root folder
- [x] macOS: Also delete all *.dmg files in `build/macOS` before running macdeployqt


bugwelle:
TODO:

Packaging scripts work for:

- [ ] Windows (mxe)
- [ ] macOS
- [ ] Linux (ApImage)
- [ ] Linux (deb)
- [ ] Linux (launchpad)

Issue #834, opened 2019-11-16
Add TVmaze as scraper


psonnosp:
**Is your feature request related to a problem? Please describe.**
The program is vulnerable when it rely on TheTVDB as its main source since that site is so unreliable, they've been messing with their new site and API for several years now and they still haven't figured it out. 

**Describe the solution you'd like**
Add TVmaze as an option under TV Scraper along TheTVDB and IMDb.

**Describe alternatives you've considered**
Adding stuff manually but I got that crazy idea out of my head rather quickly ;)

**Additional context**
https://www.tvmaze.com/blogs/21/an-alternative-when-the-thetvdb-api-is-down

bugwelle:
Hi,

thank you. TVmaze is definitely an option that I will consider. Thank you. :-)

bugwelle:
After a year I've finally rewritten our TV scraper interface. Adding TV maze is now rather easy (they have a well structured API at https://www.tvmaze.com/api ).

bugwelle:
#1096 is a first step for adding TVmaze as a provider. :-)

bugwelle:
TVmaze will be added by #1098 😄

PR #1096, opened 2020-12-17
[TvShow] Add TVmaze ID to show

bugwelle:
See #834


bugwelle:
Screen Shot 2020-12-17 at 17 37 29

PR #1133, opened 2020-12-27
[TvShowEpisode] Disable thumbnail download if not wanted 

oellan:
* Add field, getter and setter for wanting to download thumbnail (`m_want_thumbnail_download`)
 * Add detection of `EpisodeScraperInfo::Thumbnail` in `copyDetailsToEpisode` to check if thumbnail was selected in episode scraper settings
 * Fix IMDB scraper's CMakeLists.txt (for case-sensitive filesystem)
 * Add CLion-specific folders in .gitignore

### DIFF
--- a/src/scrapers/tv_show/ShowMerger.cpp
+++ b/src/scrapers/tv_show/ShowMerger.cpp
@@ -224,7 +224,9 @@ void copyDetailsToShow(TvShow& target, TvShow& source, const QSet<ShowScraperInf
 
 void copyDetailsToEpisode(TvShowEpisode& target, const TvShowEpisode& source, const QSet<EpisodeScraperInfo>& details)
 {
-    target.setWantThumbnailDownload(details.contains(EpisodeScraperInfo::Thumbnail));
+    if (details.contains(EpisodeScraperInfo::Thumbnail)){
+        target.setWantThumbnailDownload(true);
+    }
     for (EpisodeScraperInfo detail : details) {
         copyDetailToEpisode(target, source, detail);
     }
--- a/src/scrapers/tv_show/ShowMerger.cpp
+++ b/src/scrapers/tv_show/ShowMerger.cpp
@@ -224,13 +224,10 @@ void copyDetailsToShow(TvShow& target, TvShow& source, const QSet<ShowScraperInf
 
 void copyDetailsToEpisode(TvShowEpisode& target, const TvShowEpisode& source, const QSet<EpisodeScraperInfo>& details)
 {
-    bool hasThumbnail = false;
+    target.setWantThumbnailDownload(details.contains(EpisodeScraperInfo::Thumbnail));
     for (EpisodeScraperInfo detail : details) {
-        if (detail == EpisodeScraperInfo::Thumbnail)
-            hasThumbnail = true;
         copyDetailToEpisode(target, source, detail);
     }
-  target.setWantThumbnailDownload(hasThumbnail);
 }
 
 void copyDetailsToShowEpisodes(TvShow& target,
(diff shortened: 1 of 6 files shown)


src/tv_shows/TvShowEpisode.h
```diff
@@ -124,6 +125,7 @@ class TvShowEpisode final : public QObject
     void setDatabaseId(int id);
     void setSyncNeeded(bool syncNeeded);
     void setIsDummy(bool dummy);
+  void setWantThumbnailDownload(bool);
```
bugwelle, line 128:
```suggestion
    void setWantThumbnailDownload(bool wantThumbnail);
```

bugwelle (commented):
Hi,

first of all: Thank you for this pull request! 🎉 👍 

The thumbnail download is currently really intertwined with the GUI... I want to refactor that in the future but at the moment I don't have time for that. So this PR is fine by me, even if I dislike the idea of yet another boolean flag.

I will have to test this PR (didn't have time for it, yet). But if it seems to work then I'll merge this PR.

(Btw: Thanks for the CMake fix. I always forget that macOS has a case-insensitive filesystem) 

Thank you! 😄

src/tv_shows/TvShowEpisode.h
oellan: Thanks for the review.
I don't really like how I did it too, with a boolean flag, but I have a limited knowledge of QT and its type manipulation.
Maybe signals could be a better solution but with the heavy interaction between the thumbnail download code and the UI one, I didn't test this possibility.
I ran some tests and seemed to run correctly but I'll gladly rewrite some part if needed.

bugwelle: Thanks. I'm currently working on removing the GUI dependency from MediaElch's scrapers. That was a lot of work for the TV scrapers as well... No idea when I'll have time for this but just leave it as it is. :-)

src/scrapers/tv_show/ShowMerger.cpp
```diff
@@ -224,9 +224,13 @@ void copyDetailsToShow(TvShow& target, TvShow& source, const QSet<ShowScraperInf
 
 void copyDetailsToEpisode(TvShowEpisode& target, const TvShowEpisode& source, const QSet<EpisodeScraperInfo>& details)
 {
+    bool hasThumbnail = false;
     for (EpisodeScraperInfo detail : details) {
+        if (detail == EpisodeScraperInfo::Thumbnail)
+            hasThumbnail = true;
         copyDetailToEpisode(target, source, detail);
     }
+  target.setWantThumbnailDownload(hasThumbnail);
```
bugwelle, line 233:
```suggestion
    target.setWantThumbnailDownload(hasThumbnail);
```

src/tv_shows/TvShowEpisode.cpp
```diff
@@ -893,6 +893,16 @@ bool TvShowEpisode::isDummy() const
...
 }
 
+void TvShowEpisode::setWantThumbnailDownload(bool wantThumbnail)
+{
+    m_wantThumbnailDownload = wantThumbnail;
+}
+
+bool TvShowEpisode::wantThumbnailDownload() const
+{
+  return m_wantThumbnailDownload;
```
bugwelle, line 903:
```suggestion
    return m_wantThumbnailDownload;
```

src/scrapers/tv_show/ShowMerger.cpp
bugwelle: Btw. You could avoid the check in the loop by using `details.contains(EpisodeScraperInfo::Thumbnail)` :-)

bugwelle (approved):
Just two more places to format. Tested locally and seems to work fine. :-)

bugwelle (approved):

bugwelle:
Thanks! :-)

oellan:
You're welcome.
It's been a pleasure to submit a fix for a software I use and like.

PR #1141, opened 2020-12-30
[Kodi] Add "generator" tag to NFO files

bugwelle:
A `<generator>` tag contains meta details about the tools used to
generate the NFO file.  It is an unofficial tag for Kodi that may be
useful for debugging purposes.

XML looks like:

```xml
    <generator>
        <appname>MediaElch</appname>
        <appversion>2.8.3-dev</appversion>
        <datetime>2020-12-30T13:32:36Z</datetime>
    </generator>
```

`<kodiversion>` is not yet included but may come at a later point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@
 
 ### Changes
 
- - *tbd*
+ - NFO files now contain a `<generator>` tag which holds meta information about MediaElch.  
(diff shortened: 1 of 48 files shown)


bugwelle:
@KarellenX Just FYI. I hope this tag may help users on the Kodi forum in the future. Just to show which version of MediaElch generated the NFO. :-)

KarellenX:
Yes that should be helpful, and Kodi will just ignore it so no problems. Thanks

Issue #1139, opened 2020-12-29
Movie/TV Show Year Autotag


GarfieldS:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I would like to export my current list of scrapes with TV/Movie Year to make categorizing and filtering media by year very simple. 

**Describe alternatives you've considered**
I looked to see if there was an export feature that could export the info to an Excel compatible format but did not find any that would also allow me to choose attributes. Potentially other software could handle this but wanted to see if it was possible withing the program itself. There is a date for when first aired and that attribute is good enough because in Excel it would be easy to get the year.

**Additional context**
I create an online searchable TV/Movie Portal and would like to add this attribute as another way for users to find content. Eventually, I would love to allow searchers for 80's, 70's, 90's, etc.

ticao2:
> I would like to export my current list of scrapes with TV/Movie Year  

I don't know if this will help you, but ...
I needed to export my movie collection to an HTML table and an Excel spreadsheet.
And I created 5 themes for exporting data.
The spreadsheet theme, in .csv, is attached here.
Just place the **T_csv10** folder inside the **export_themes** folder that is in the **MediaElch.exe** folder.
When you open your MediaElch, it will be one of the export options.

[T_csv10.zip](https://github.com/Komet/MediaElch/files/5752583/T_csv10.zip)
The text is in my language but if you think you can solve your problem I can put everything in English.

GarfieldS:
> The spreadsheet theme, in .csv, is attached here.

Thank you very much. I will test

GarfieldS:
I was unable to find an export_themes folder

ticao2:
> 
> I was unable to find an export_themes folder

You can create the folder yourself, but I think the best solution, the safest, will be the following.
The **export_themes** folder is automatically created when you install any of the themes that come with MediaElch.
Then go to **Settings > Export**
You will see the list with **4 themes** that can be installed.
**DigitalRabbit - Datatables**, **Dummy Theme**, **Hidaya Theme** and **MediaElch**.
I suggest installing the **Dummy Theme**. It may be useful in the future.
Then click on its green button and it will be installed inside the **export_themes** folder that will be automatically created inside the MediaElch.exe folder.
I believe that it is now better to **close MediaElch**.
Copy the folder I sent into this newly created folder.
The folder name, **T_csv10**, cannot be changed.
Then open MediaElch and go to **Settings > Export** again.
[...]

PR #1147, opened 2021-01-01
[Export] Add CSV export functionality for movies

bugwelle:
Add simple CSV export functionality. It comes with a (for now) hidden
dialog box that can be opened using Ctrl+Shift+E.

Only movies are supported as of now. More may come.

Screen Shot 2021-01-01 at 14 24 12

See #1139


bugwelle:
Btw, the CSV has a BOM ( https://de.wikipedia.org/wiki/Byte_Order_Mark ). Without it, Excel will not recognize UTF-8 on import.

PR #1161, opened 2021-01-05
[CsvExport] Improve export dialog, add more export types

bugwelle:
Screen Shot 2021-01-05 at 14 15 37

Feedback is welcome. :-)


bugwelle:
FYI @GarfieldS , @ticao2 

Improved the export dialog. What's left is that the item order and checkboxes are stored through restarts (i.e. is they are checked) .

Issue #1162, opened 2021-01-06
I'd really like to see Tags in the Custom Movie Scrapper.


Headspace256:
I have to scrape every movie twice, once with imbd to get tags then with Custom scrapper or TMDB to get sets. 

It'd be great if the custom scrapper gave you the option to scrap tags from IMBD as it is already getting other info from IMBD

Thanks for a great application and taking the time to read this.

bugwelle:
Hi,

a rewrite of movie scrapers is planned. This includes the custom movie scraper. ETA: ~next 2 months (hopefully, no promises).

Headspace256:
> Hi,
> 
> a rewrite of movie scrapers is planned. This includes the custom movie scraper. ETA: ~next 2 months (hopefully, no promises).

Thanks that's great, and I promise I won't hold you to it.

bugwelle:
> a rewrite of movie scrapers is planned. This includes the custom movie scraper. ETA: ~next 2 months (hopefully, no promises).

Well, it was more like 2 years. But I finally got to rewrite the movie scrapers. There have been _many_ bugs I found and had to fix before I could rewrite it. 

Now you can select any scraper for tags and it will be used. See https://github.com/Komet/MediaElch/pull/1571
(the change is only so small, because the refactoring of code made it easy)

PR #1163, opened 2021-01-07
[MovieScraper] Refactor and move to job based scrapers

bugwelle:
Just like for TV scrapers. This will allow us to have parallel scraping, parallel searches and fewer race conditions that would be possible if we move scraping to a background process (like changing the scraper's language in the settings during scraping).

Should fix:
 - #1162


bugwelle:
TODO:

- [ ] Rework CustomMovieScraper
- [ ] Rework CustomMovieScraper handling in MovieDialog -> in a scraper needs search first, let the user select correct movie

bugwelle:
Ok, so far I think I've implemented everything. Now I need to do some testing:

- [ ] Add scraper tests
- [ ] Manual test
   - [ ] Single movie scraping
      - [ ] Load all details
         - [ ] from the same scraper
         - [ ] from different scrapers
      - [ ] Load no details (via settings)
   - [ ] Multi movie scraping
      - [ ] Load all details
         - [ ] from the same scraper
         - [ ] from different scrapers
      - [ ] Load no details (via settings)

PR #1170, opened 2021-01-09
.desktop-file add categories

Stunkymonkey:
`Database` because you can update the Kodi-Database


bugwelle:
Interesting. https://specifications.freedesktop.org/menu-spec/latest/apas02.html actually lists `Qt` as a category... 😮

bugwelle (approved):

PR #1184, opened 2021-01-15
[i18n] Update translation release binaries

bugwelle:
The update script failed to run "lrelease" of Qt's linguistic tool.

@ticao2  FYI :)


bugwelle:
Reason. `lrelease` silently failed... `--version` gives:

```
Passing .pro files to lrelease is deprecated.
Please use the lrelease-pro tool instead, or use qmake's lrelease.prf
feature.
```

Ok... https://doc.qt.io/qt-5/linguist-manager.html isn't updated yet.

I've updated our i18n update script.

Issue #1176, opened 2021-01-11
Questions / Issues for TV Shows in v2.8.4


ticao2:
**Is your feature request related to a problem? Please describe.**

**Doubts and Facts**
If you wish I can issue the item individually.

**Reminder:** I haven't used it in a long time. Since 2015.
So I don't remember if it was always like this.
So I'll just see.

**SERIES (TV Show) 2.8.4**

**1 Filter** = only works for Title, specifically Local Title.

**2 Make Scrapper** = In Movies I can do with the year in the title. In Series not.
Example: **Ad Astra (2019)** versus **Cosmos (2014)**

**3 TAGs** = is the keywords of TMDb, I presume.

**4 Advancedsettings** = Genres are not mapped. Certification is not mapped.

**5 Certification** = Does not display dropdown with those of the other Series.

**6 TV Tune** = What is it?

**7 Original Title** = is not in the UI.

bugwelle:
Hi,

I'll try my best to answer your questions:

1. Yes. At the moment only the title is supported. Please let me know what other fields you would like. I wanted to refactor the filter because it has a few coding issues (like memory leaks, etc). That's why I don't want to spend too much time adding fields at the moment. :)
2. Could you elaborate what you mean? :)
3. Yes. TMDb for TV shows should use a show's keywords as tags. Needs documentation. :)
4. Ohhh...  genres should be mapped for TheTvDb at least. TMDb and TVmaze are missing, that is true. I'll look into it.
5. True. Would be a nice feature. Could you open a feature request for it? :)
6. See https://mediaelch.github.io/mediaelch-doc/tvshow/index.html#tv-tune :-)
7. AFAIK we don't support originaltitle for TV shows, yet, do we? 
   Please open a new feature request for it. :-)

ticao2:
I will try to be concise. **:-)**

**1 Filter** = I divided my film library into 3 basic areas:
    Series = Big Bang, House ...
    Mini Series = Band of Brothers, Taken (2002) ...
    Doc Series Mini = Cosmos (1980), Cosmos (2014), Richard Feynman (1964) The Character of Physical Law ...
I use TAGs and/or Genres to make the distinction.
It would be interesting to be able to view the separate groups.
**2 Make Scrapper** = It's just my curiosity.
All titles in my movie and series folders are in the following format, for example:
**Title_pt-BR (YEAR) mkv @ 1080p 6.4G (Dual)**
[...]

PR #1186, opened 2021-01-16
[Filter] Add genre and tag TV show filters

bugwelle:
See https://github.com/Komet/MediaElch/issues/1176


bugwelle:
Ok, this needs more time. Current issues:

1. What model does the `QModelIndex` point to?
2. How to handle parent/child ralationships?
3. When to update the view?

bugwelle:
Closing: Won't fix

If I implement it the way this PR does, MediaElch's filter will be extremely slow. The current filter is limited and it will remain this way until it is rewritten sometime in the future.

PR #1207, opened 2021-01-28
[CMake/QMake] Generating *.qm files on demand

bugwelle:
Don't store binary files in the Git repository but instead generate
them by using QMake or CMake.

TODO:

- Check if translations get deployed (translations are deployed if MediaElch->About->Close-button is translated)
  - [ ] Windows
  - [x] macOS
  - [ ] Linux AppImage


bugwelle:
For packaging script, see #1084

PR #1239, opened 2021-02-28
[CustomTvScraper] Add "Do not use" field

### LINKED ISSUE #1237: Option to not use Rating for TV Shows.
In the configuration of the Film Scraper, there is a way to stop using the Rating, just changing the selection. In TV Shows this option does not exist, I believe this change is easy, hugs ..

bugwelle:
Fix #1237

-------------

TODO:

1. Avoid image loading
2. Do not overwrite existing details


bugwelle:
Discarding. The idea is good, but the UI would be cluttered.

PR #1247, opened 2021-03-06
[Project] More changes towards Qt6 compatibility 

bugwelle:
This commit has some fixes for Qt6.  Some deprecated enums/methods
were replaced.  Comparisons between QStringView and char* required
either QStringView on the right-hand-side or QLatin1String.  We use
the latter now as that's supported on older Qt versions as well.

Some unused `QRegEx` includes were removed.

Some conditionals were used, e.g. for method calls that are required
in Qt5 but are the default (and therefore removed) in Qt6.

This commit does NOT make MediaElch compatible with Qt6.
We still use QRegEx in a lot of places.  Furthermore, we have to wait
for Qt 6.2 to finally get QMultiMediaWidgets.

### DIFF
--- a/src/settings/Settings.cpp
+++ b/src/settings/Settings.cpp
@@ -13,6 +13,7 @@
 #include <QDesktopServices>
 #include <QMutex>
 #include <QMutexLocker>
+#include <QStandardPaths>
 
 static constexpr char KEY_ALL_DATA_FILES[] = "AllDataFiles";
 static constexpr char KEY_AUTO_LOAD_STREAM_DETAILS[] = "AutoLoadStreamDetails";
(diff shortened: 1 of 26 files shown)


src/settings/Settings.cpp
```diff
@@ -1328,15 +1329,15 @@ mediaelch::DirectoryPath Settings::databaseDir()
     if (advanced()->portableMode()) {
         return mediaelch::DirectoryPath(applicationDir());
     }
-    return mediaelch::DirectoryPath(QStandardPaths::writableLocation(QStandardPaths::DataLocation));
+    return mediaelch::DirectoryPath(QStandardPaths::writableLocation(QStandardPaths::AppLocalDataLocation));
```
bugwelle, line 1332: `DataLocation` and `AppLocalDataLocation` are the same. :)

PR #1248, opened 2021-03-06
[Preview] Qt6 compatibility changes

bugwelle:
This PR is heavily WIP.

-------------

Qt 6.1:

Qt 6.1 does not have QMultiMediaWidgets, so I simply removed all functionality that requires it.
Furthermore, we have a lot of QRegEx, so I commented out that code as well.

Some other changes were required as well. I didn't look into issues with CMakeLists.txt too much.
I simply added required modules until it compiled. I need to split MediaElch into actual libraries (we currently use object libraries).

Also, QuaZip requires the compatibility module in v1. As I didn't want to use it, I've simply removed the QuaZip code for now.

Screen Shot 2021-03-06 at 17 06 00

-------------

Qt 6.2

Multimedia was added but we have issues with QuaZip which relies on QTextCodec...

### DIFF
--- a/src/export/ExportTemplateLoader.cpp
+++ b/src/export/ExportTemplateLoader.cpp
@@ -6,9 +6,6 @@
 #include <QNetworkReply>
 #include <QXmlStreamReader>
 
-#include "quazip/quazip.h"
-#include "quazip/quazipfile.h"
-
 #include "globals/VersionInfo.h"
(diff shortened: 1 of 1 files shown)


bugwelle:
We're getting there! :tada:

See #1346

bugwelle:
Replaced by #1366

PR #1282, opened 2021-04-12
[Packaging] Fix copyright string "LGPL-2.1" -> "LGPL-3.0"

bugwelle:
Our file "COPYRIGHT" has been LGPL-3 for a long time.
For some reason, the copyright files in our packaging files for
Debian and openSUSE were wrong.

As far as I can see, we use "LGPL-3.0-only".

-------------

@Komet I need approval from your side. As per mail, LGPLv3 is correct but I'm unsure about the "-only" suffix. :)


Komet:
👍

Komet (approved):

PR #1291, opened 2021-04-23
[File] Add Filesystem class for better testing

bugwelle:


bugwelle:
The idea was to pass an explicit file system object to each class that handles files. However, I now think that having global functions or at least a singleton isn't bad in _this case_. Simply because we're not a library, we're an app. Having a central place that handles files is _useful_ here.

bugwelle:
Further ideas: 

- Should there be a central object that others can connect to?
- Should we leave it at the current `installFileSystemLog` state? We can move to the before mentioned object quickly by adding it to `FileSystemLog.cpp`

bugwelle:
Disgarding.

PR #1317, opened 2021-05-06
[MovieFileSearcher] Copy "searcher" data QApplication::processEvents(); 

### LINKED ISSUE #1316: Crash when reloading movies Library
**Describe the bug**
Every time I reload the Movies library, mediaelch crashes

**To Reproduce**
1. Go to ‚Movies.'
2. Click on ‚Reload.‘
3. See Crash

**MediaElch Version:**
 - [ ] 2.8.11-dev (nightly)
 - [x] 2.8.10 stable

**Operating System:**
 - [ ] Windows
 - [x] macOS

**Additional context**
I’ve added [macOS Crash Report.txt](https://github.com/Komet/MediaElch/files/6433990/macOS.Crash.Report.txt)

bugwelle:
It could happen that we run multiple times into `onDirectoryLoaded()`
at once because of the call to `QApplication::processEvents();`.

This also meant that it could happen that the searchers are already
deleted before we return to the normal control flow in the first
call to `onDirectoryLoaded()`.  All references to `searcher` after
`QApplication::processEvents();` were therefore use-after-free bugs.

-----------

See #1316, #1315

### DIFF
--- a/.ci/ci_utils.sh
+++ b/.ci/ci_utils.sh
@@ -56,7 +56,8 @@ gather_project_and_system_details() {
 	GIT_HASH=$(git --git-dir=".git" show --no-patch --pretty="%h")
 	DATE_HASH=$(date -u +"%Y-%m-%d_%H-%M")
 	DATE_DESC=$(date -u +"%Y-%m-%d %H:%M")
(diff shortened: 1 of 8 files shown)


bugwelle:
Even though this may not fix the issue referenced above, it _does_ fix a few issues. I'll merge it.

PR #1320, opened 2021-05-08
[MovieFileSearcher] Check if aborted after processEvents() call

bugwelle:
For #1315


src/movies/file_searcher/MovieFileSearcher.cpp
```diff
@@ -162,7 +162,7 @@ void MovieFileSearcher::onDirectoryLoaded(MovieDirectorySearcher* searcher)
 
         if (i % 40 == 0 && i > 0) {
             // Each call to processEvents() could potentially have triggered abort();
-            QApplication::processEvents();
+            // TEST: QApplication::processEvents();
```
bugwelle, line 165: I _assume_ that this call is problematic.

PR #1330, opened 2021-05-16
[NameFormatter] Use singleton again

bugwelle:
There is only one instance of this class (singleton).  While I
mostly dislike singletons, in this case there are reasons.  If we
do not use a singleton and do not store QRegularExpression objects,
each movie that is loaded has to instantiate all QRegularExpressions
for the "exclude word list".  Due to JIT compilation, this has
a _huge_ time disadvantage.  In some tests, 50% of the time went
into the NameFormatter prior to using a singleton.
See #1229 for more details.


bugwelle:
Also a time improvement:

After name formatter change:  2048ms
Before name formatter change: 2564ms

PR #1366, opened 2021-11-17
[QuaZip] Update submodule to QuaZip 1.1 and add basic support for Qt6

bugwelle:
Update submodule to QuaZip 1.1.  Furthermore, apply fixes so that we
can also work with QuaZip 1.1 for Qt6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@
  - CMake: Option `ENABLE_TESTS` was added. Without it, MediaElch's tests will
    not be build.  The reason is that normal users, that only want to build
    MediaElch, do not need them.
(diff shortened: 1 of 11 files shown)


bugwelle:
Tested on Fedora 35 which already ships Qt6. Works with and without `-DUSE_EXTERN_QUAZIP`.

And if the external QuaZip can't be found, we get:

```
CMake Error at CMakeLists.txt:224 (message):
  Could not find QuaZip 1.x using CMake.  And can't use quazip5 as Qt 6 was
  requested!
```

:tada:

Notes: `-DCMAKE_CXX_FLAGS="-fpic"` was required for Fedora.

All that's left is to remove Qt Quick Controls 1

Issue #1378, opened 2022-01-01
Allow season names to be used when renaming season folders


ReK42:
Some shows have named seasons. MediaElch supports these names and will save them in the NFO file, but it cannot be used when automatically renaming the season folders. A useful feature would be to implement the <seasonName> variable in the renamer. For example, the settings in the first screenshot below should result in the folder structure shown in the second screenshot.

![image](https://user-images.githubusercontent.com/4775937/147843431-01c52e70-7904-4349-8445-cfd8c4ffa40f.png)

![image](https://user-images.githubusercontent.com/4775937/147843452-b94de623-8883-4c3f-9afa-49b01742fe9d.png)

bugwelle:
This was implemented by @oellan in  #1381. :-)

You can now use `<seasonName>` and `{seasonName}` in the renamer. The former is the placeholder, the latter is the condition.

Regards,
Andre

bugwelle:
A new nightly for Linux is now available. One for Windows will be uploaded in the next 10min.

PR #1381, opened 2022-01-08
[Renamer] Add "seasonName" placeholder and condition

oellan:
The placeholder holds the season name. If the season name isn't specified, it holds an empty string.
The condition is valid if the season name is an empty string.

Implements #1378


bugwelle (approved):
LGTM

Thank you very much for your pull request. I very much appreciate it.  :smile:



Regards,
Andre

PR #1384, opened 2022-01-18
[Renamer/Import] Interchange originalTitle on replace

roalter:
When originalTitle is empty fall back to title for the placeholder replacement.
Otherwise you can't rely on renaming to <originalTitle>; biggest problem on direct
Bulk Importing Files.

Additional *always* update the lists after import


bugwelle:
Thank you very much for this PR. Could you explain why this is the case:

> Additional always update the lists after import

? :-)

bugwelle:
Hi,

I've taken your approach an MediaElch will now use a movie's title as the fallback if the originalTitle does not exist.
This is part of  #1437

But still, could you tell me why the list should _always_ be updated after an import? :smile: 

Regards,
Andre

bugwelle:
Closing due to inactivity.

PR #1421, opened 2022-02-23
Add more languages to the sort tokens

somini:
Sourced from here: https://en.wikipedia.org/wiki/Article_(grammar)#Tables 

Going over the list and including the rest (without major coflicts) is left as an exercise to the reader.

This is *much* more complex than this simple feature can handle, but these should cover [~2 billion people](https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers).

### DIFF
--- a/docs/advancedsettings.xml
+++ b/docs/advancedsettings.xml
@@ -71,16 +71,49 @@
         ("The Movie" -> "Movie, The").
     -->
     <sorttokens>
-        <token>Der</token>
-        <token>Die</token>
-        <token>Das</token>
-        <token>The</token>
-        <token>Le</token>
-        <token>La</token>
-        <token>Les</token>
-        <token>Un</token>
-        <token>Une</token>
-        <token>Des</token>
+		<!-- English -->
+		<token>The</token>
+		<token>A</token>
+		<token>An</token>
+		<!-- German -->
+		<token>Der</token>
+		<token>Die</token>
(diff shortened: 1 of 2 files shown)


docs/advancedsettings.xml
```diff
@@ -71,16 +71,49 @@
...
+		<token>A</token>
+		<token>An</token>
+		<!-- German -->
+		<token>Der</token>
+		<token>Die</token>
+		<token>Das</token>
+		<token>Des</token>
+		<token>Dem</token>
+		<token>Den</token>
+        <token>Ein</token>
```
bugwelle, line 85:
```suggestion
         <token>Ein</token>
```

bugwelle (commented):
Thank you very much! I'm fine with this change. Please note that these have to be mirrored in AdvancedSettings.cpp on line 11 (`m_sortTokens`). `advancedsettings.xml` reflects the default values from `AdvancedSettings.xml` :-)

Shall I add them later or do you want to?

bugwelle:
I updated the formatting and `AdvancedSettings.xml`. Thank you for this improvement. :)

bugwelle:
The `advancesettings.xml` file is just a file I put together to have all the defaults from `AdvancedSettings.xml` in a Copy&Pastable file. :-)

Also, this is what code-review is for. You can't know everything :smile:

Thank you for your contribution to this project!

PR #1436, opened 2022-04-10
[Renamer] Mark movie "changed" if subtitle was renamed

### LINKED ISSUE #1432: Files are renamed, but old filename stays in nfo
**Describe the bug**
When using the file renamer to rename for instance srt files, the name for this file in the nfo stays the old one. But that file is gone. 

**To Reproduce**
1. Select a movie that has an srt file
2. Open file renamer
3. Change filename pattern to `<title>.<year>.<extension>`
4. Rename
5. Check the srt filename in the nfo file

**Expected behavior**
When renaming files, any file that is mentioned in the nfo file should also be renamed

**MediaElch Version:**
 - [ ] 2.8.17-dev (nightly)
 - [X] 2.8.16 stable

**Operating System:**
 - [ ] Windows
 - [ ] macOS
 - [X] Linux (distribution: Ubuntu.18.04)

bugwelle:
Fix #1432

------------

[StreamDetails] Refactor "hasStreamDetails" handling

This commit refactors the stream details handling a bit. Previously,
we used a member variable for `streamDetailsLoaded()` that tracked
whether stream details were loaded.  However, this was not reliably
set throughout our code base.  We now simply use the StreamDetails
class for tracking this state.

------------

[Renamer] Mark movie "changed" if subtitle was renamed

If a movie's subtitles are renamed, the movie must be marked as
"changed" or otherwise the `<subtitle>` tag in its NFO file won't be
updated, or rather: The user won't notice that the tag is not updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
  - Multi-Episode numbers such as `S01E01E02E03.mov` are now correctly identified (#1429)
  - UI: The "subtitle" column of movies is now also green if there are external subtitles (#1435)
+ - Renamer: If subtitles are renamed, movie is marked "changed" (#1432)
+   This is necessary to update the `<subtitle>` tag in the NFO file.
 
 ### Changes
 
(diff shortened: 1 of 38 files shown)


bugwelle:
Screenshots (in German)

Screenshot_20220410_164946
Screenshot_20220410_165006
Screenshot_20220410_165016
Screenshot_20220410_165030

PR #1454, opened 2022-07-27
[TmdbApi] Use "aggregate_credits" call for shows, Prevent invalid actor profile URLs

TheTyrius:
Hello,

please excuse if this PR should have been a discussion issue first, I am hoping this way it is easiest to see what I am trying to achieve.

I noticed that an actor was missing from the metadata of a tv show scraped by MediaElch using Tmdb.
The show in this case was one of the main characters in Stargate SG1 (https://www.themoviedb.org/tv/4629-stargate-sg-1).
On the web "Richard Dean Anderson - Jack O'Neill" is listed first, but he was missing from the saved metadata.

An investigation showed that the "credits" API call defaults to the cast of the last season of a show. Some time later an "aggregate_credits" call has been added that returns the full cast more alike it is shown on the website.

I would like to switch to this call as to me it seems unexpected to be limited to the last season (which in this case omitted a main character of the overall show).
Apologies if this has been discussed before, I did not find mentions of these calls. What is your opinion on this change?

I also noticed that "profile_path" can be null (in both endpoints), in this case an incomplete/invalid thumb-URL would be saved.

Best regards,
Philipp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
  - HotMovies: Search and scraping of ratings was fixed
  - Fixed crash if custom movie scraper was used with adult scrapers
  - Kodi NFO files will only contain `<uniqueid>` if there are valid ones
+ - TMDb TV now scrapes the full cast of a TV show instead of only the last season's (#1454)
 
 ### Changes
 
--- a/src/scrapers/tmdb/TmdbApi.cpp
+++ b/src/scrapers/tmdb/TmdbApi.cpp
@@ -180,7 +180,7 @@ QUrl TmdbApi::getShowUrl(const TmdbId& id, const Locale& locale, bool onlyBasicD
     QUrlQuery queries;
     if (!onlyBasicDetails) {
         // Instead of multiple HTTP requests, use just one for everything.
-        queries.addQueryItem("append_to_response", "content_ratings,keywords,external_ids,images,credits");
+        queries.addQueryItem("append_to_response", "content_ratings,keywords,external_ids,images,aggregate_credits");
     }
     return makeApiUrl(QStringLiteral("/tv/") + id.toString(), locale, queries);
 }
--- a/src/scrapers/tv_show/tmdb/TmdbTvShowParser.cpp
+++ b/src/scrapers/tv_show/tmdb/TmdbTvShowParser.cpp
@@ -190,7 +190,11 @@ void TmdbTvShowParser::parseInfos(const QJsonDocument& json, const Locale& local
             actor.name = actorObj["name"].toString();
             actor.role = roles.join(", ");
             actor.id = QString::number(actorObj["id"].toInt());
-            actor.thumb = m_api.makeImageUrl(actorObj["profile_path"].toString()).toString();
+
+            if (!actorObj["profile_path"].isNull()) {
+                actor.thumb = m_api.makeImageUrl(actorObj["profile_path"].toString()).toString();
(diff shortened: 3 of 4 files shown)


TheTyrius:
Ups I forgot to mention: In this case an actor can be credited with multiple roles throughout the show. I opted for a comma separated list in the order returned by the API, but if there is a better mechanism for this in place please advice :)

bugwelle:
Hi there,

another SG1 fan 👍 

Thank you very much for this PR! Opening a PR directly for small changes such as this one is totally fine!

> An investigation showed that the "credits" API call defaults to the cast of the last season of a show.

Wow... Thanks for investigating!

Your code looks fine. Nevertheless, I'll test it locally when I'm home (this evening I'd say).

I'll also add a scraper test to ensure that this won't happen again (e.g. through an accidental change) ( + changelog entry). 

Regards,
Andre

bugwelle (approved):

TheTyrius:
> another SG1 fan +1

For sure, can't have Jack missing ;) Plus this should affect any show with a cast-change.
> Wow... Thanks for investigating!

I got the hint from [forum posts](https://www.themoviedb.org/talk/5de3116fa7e363001969bed4), sadly the "credits" api-docs don't say so themselves (only the [aggregate_credits](https://developers.themoviedb.org/3/tv/get-tv-aggregate-credits) one confirms it in its description)
> Your code looks fine. Nevertheless, I'll test it locally when I'm home (this evening I'd say).

Please do, I only just cloned the code after being a user for quite some time and don't know much about the project as a whole.
I also didn't do _much_ additional testing tbh, as so far I can't even get the tests to compile/execute locally (`ninja: error: unknown target 'test'`, following your testing.md guide on linux)
> I'll also add a scraper test to ensure that this won't happen again (e.g. through an accidental change) ( + changelog entry).

Thank you. I'm fully aware that this change is not without trade-offs: First I'm not sure how or even if the "roles" are ordered. And second, instead of a (too) short main cast we now get the _full_ cast, which is a long list in this case, but at least ordered (by episode count I guess).
To me this would be the expected behaviour for a tv show. Looking at my library, movies seem to get more than just the main cast as well already. Ultimately it is for you to judge once you can have a look, certainly there is no rush at all :)

TheTyrius:
> I can't even get the tests to compile/execute locally (`ninja: error: unknown target 'test'`, following your testing.md guide on linux)

I passed `-DENABLE_TESTS=true` to cmake to enable the tests. As far as I can tell this is a recent change that is not yet reflected in the testing.md document :)

bugwelle:
> I passed `-DENABLE_TESTS=true` to cmake to enable the tests. As far as I can tell this is a recent change that is not yet reflected in the testing.md document :)

Arghh... Yes! Thanks. I disabled the tests per default as most users don't need them. I totally forgot about `-DENABLE_TESTS=true`!

bugwelle:
Guideline adapted with edee4131ddf3a0b1892b1ac5b58cd262d68bba40

bugwelle:
I've added tests that fail with `credits` but work with `aggregate_credits`.

bugwelle:
Merged. Thank you very much. :-)

Even though 680 is quite a lot. Maybe that should be a config option. We'll see if someone asks. :)

TheTyrius:
It is 😅 making it configurable was simply beyond my possibilities at this point. I hope it will be fine.
Thank you for your time and work on MediaElch!

Issue #1475, opened 2022-10-04
Crashes on poster select


warchild13:
When a film is selected double click on poster, alternate poster window opens. Select alt poster and ME Crashes instantly.
Holds true on last few builds.

Debian/GNU 11
QT 5.15.2
Kernel 5.10-18-amd64

bugwelle:
Reproduced. Relevant crash log:

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   QtNetwork                     	       0x1090eb1a8 QNetworkReply::isFinished() const + 4 (qnetworkreply.cpp:590) [inlined]
1   QtNetwork                     	       0x1090eb1a8 QNetworkReply::isRunning() const + 8 (qnetworkreply.cpp:603)
2   MediaElch                     	       0x1079b0cc0 ImageDialog::cancelDownloads() + 96 (ImageDialog.cpp:498)
3   MediaElch                     	       0x1079b0c49 ImageDialog::accept() + 25 (ImageDialog.cpp:170)
```

Seems c25691b7321cd936d4e045e0fea133a87321ff1f was not enough.

Thanks for reporting!

bugwelle:
Could you test #1476 ? :-)

For some reason, I couldn't reproduce this bug with address sanitizer enabled.

warchild13:
That fixed it!
Thanks. For me MediaElch is used daily indispensable!

bugwelle:
> That fixed it!

Great! Thanks for letting me know! Thank you very much for reporting this bug!

> Thanks. For me MediaElch is used daily indispensable!

That is great to hear! Please let me know if you find any other crash/bug. If I've got time, I will try to fix it ASAP. :-)

PR #1476, opened 2022-10-04
fix(ImageDialog): Clear pointer after call to deleteLater()

bugwelle:
This (hopefully) fixes a use after free bug, where `m_currentDownloadReply->abort()` was called after `m_currentDownloadReply` was deleted.  This went undetected, because `deleteLater()` does not clear the pointer (only if `QPointer<>` were to be used).

Maybe we should use QPointer here, but that's for another commit.

------------

For https://github.com/Komet/MediaElch/issues/1475


bugwelle:
Merging. Even if it does not fix the crash, this is definitely cleaner.

Issue #1490, opened 2022-11-06
Releases: Add Qt6 releases for macOS/Windows


bugwelle:
Will (hopefully) also fix #429

TODO:

- [x] automate Qt6 release via scripts (as we do for Qt5)
   - [x] macOS => #1491
      - [ ] change required minimum version for the Qt6 build
   - [x] Windows => https://github.com/Komet/MediaElch/pull/1492
- [x] update Download page to tell users which version to download (Windows 7 - 8: Qt5, Windows 10: Qt6)
- [ ] test it
- [x] check resulting bundle sizes
  - [x] macOS (needs hack)
  - [x] Windows (+ 3MiB in ZIP file)
- [x] Test QML file

bugwelle:
We now have scripts for everything. Closing. :)

PR #1491, opened 2022-11-06
ci(macOS): Adapt build script to build Qt5 and Qt6

bugwelle:
For  #1490


bugwelle:
Bundle size is 130MiB instead of 70MiB. Will need to look into it in another PR.

PR #1492, opened 2022-11-06
ci(Windows): Adapt MXE build scripts to build Qt5 and Qt6

bugwelle:
For #1490


bugwelle:
Image is now building on Jenkins. If it succeeds, we can test Qt6 build + packaging with MXE.

bugwelle:
Image build succeeded. I'm not building it locally. Packaging needs some work, because I have to figure out which DLLs are required.  Sadly, I'm not aware of any good automatic process, so this will be trial and error (at least Windows has some good popups).

bugwelle:
Got everything working on my Windows 10 test system. All DLLs are packed into the ZIP as well as all QML dependencies. Jenkins is currently building a Qt6 snapshot.

PR #1520, opened 2022-12-27
fix(TV show): Adapt `<episodeguide>` to Kodi v20 changes

### LINKED ISSUE #1519: TV show: Update `<episodeguide>`
See https://github.com/xbmc/xbmc/issues/17341#issuecomment-1329636946 and https://forum.kodi.tv/showthread.php?tid=370489

bugwelle:
Kodi v19.5 and later use a different episodeguide format.  Use that as well.  Since we don't need to read the episodeguide (since we already have all unique IDs), we can simply write what we already know: A map of IDs in JSON format.

Furthermore, remove `year` for Kodi v19 and later, because in v20, only `premiered` must be used.

Fixes #1519

### DIFF
--- a/test/resources/show/kodi_v20_show_Game_of_Thrones_TvDb_episode_guide.nfo
+++ b/test/resources/show/kodi_v20_show_Game_of_Thrones_TvDb_episode_guide.nfo
@@ -20,15 +20,12 @@
     <plot>Seven noble families fight for control of the mythical land of Westeros. Friction between the houses leads to full-scale war. All while a very ancient evil awakens in the farthest north. Amidst the war, a neglected military order of misfits, the Night's Watch, is all that stands between the realms of men and the icy horrors beyond.</plot>
     <mpaa>TV-MA</mpaa>
     <premiered>2011-04-17</premiered>
-    <year>2011</year>
     <dateadded>2019-07-05 17:35:08</dateadded>
     <status></status>
     <studio>HBO</studio>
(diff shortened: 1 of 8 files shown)


test/resources/show/kodi_v20_show_Game_of_Thrones_TvDb_episode_guide.nfo
```diff
@@ -20,15 +20,12 @@
...
-    <year>2011</year>
     <dateadded>2019-07-05 17:35:08</dateadded>
     <status></status>
     <studio>HBO</studio>
     <runtime>55</runtime>
     <trailer></trailer>
-    <episodeguide>
-        <url post="yes" cache="auth.json">https://api.thetvdb.com/login?{&quot;apikey&quot;:&quot;439DFEBA9D3059C6&quot;,&quot;id&quot;:121361}|Content-Type=application/json</url>
-    </episodeguide>
+    <episodeguide>{&quot;imdb&quot;:&quot;tt0944947&quot;,&quot;tvdb&quot;:&quot;121361&quot;,&quot;tvmaze&quot;:&quot;82&quot;}</episodeguide>
```
bugwelle, line 28: FYI @KarellenX : Thank you very much for notifying me. 

If Kodi v19 or later is used (i.e. the corresponding option is set in MediaElch's settings), then we create this `<episodeguide>`. The field itself is not used by MediaElch, i.e. not read, only written. :)

PR #1546, opened 2023-02-19
fix(UI): Adapt color labels for dark theme

### LINKED ISSUE #1545: In dark mode, items that have a label set are hard to see
**Describe the bug**
When using the Dark mode, any items that have a label set are almost impossible to read because the text still renders as white. 

**To Reproduce**
1. Set Dark mode
2. Assign a label to a movie (eg green)
3. The label renders in bright green with white text. Happens to all colours

**Expected behavior**
In dark mode, text in labels should render dark (or adaptive, based on label colour).

**Screenshots**
mediaelch

**MediaElch Version:**
 - [ ] 2.10.1-dev (nightly)
 - [x] 2.10.0 stable

**Operating System:**
 - [x] Windows
 - [ ] macOS

bugwelle:
Fix #1545

------

Color labels were for light theme only.  That made it nearly unreadable
in dark theme.

@ironhussar What do you think?

dark

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,9 +1,10 @@
 [main]
 host = https://www.transifex.com
 
-[mediaelch.main]
-file_filter = data/i18n/MediaElch_<lang>.ts
-minimum_perc = 20
-source_file = data/i18n/MediaElch_en.ts
-source_lang = en
-type = QT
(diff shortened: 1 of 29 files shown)


ironhussar:
That looks excellent! Toning the colours down is a great solution.

Issue #1564, opened 2023-04-29
Scraping a directory with 40 movies will display just 2 of them (but 40 is recognized)


opc241:
The scraper delivers 2 of 40 movies after reading in the directory. What is the problem here?

bugwelle:
Hi,

could you tell me a bit more about your system? See the issue template here: https://github.com/Komet/MediaElch/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

1. Which operating system do you use?
2. Do you have "separate directories" ("Sep. Verzeichnisse") enabled in your directory settings?

Regards,
Andre

opc241:
Hi Andre, 

no problem, i am using windows and yes, separate directories is enabled.

opc241:
It is a network attached drive i am using. No, the two dispayed movies are not the first one. 

Ok thanks for the hint with the debug, here is the logging file attached, maybe that helps.
[debug.log](https://github.com/Komet/MediaElch/files/11359959/debug.log)

bugwelle:
There are exactly 38 lines with `[MovieXmlReader] No <movie> tag in the document`. My guess: You have either NFO files in each movie directory or other files MediaElch thinks are NFO files. It could be that MediaElch simply ignores movies if we find an NFO file which we can't parse. But of course MediaElch _should_ show the movie, similar to when there isn't any NFO file at all.

To test this idea: Could you go into one of the movie directories that isn't recognized? If there is a `*.nfo` file, e.g. `movie.nfo`, could you paste it's contents here? If you don't want to (after all, they are _your_ movies; I don't need to know which movies you have), could you delete an NFO file if there is any and reload all movies? If it appears, then I have a very good guess what the bug is.

Anyway, thanks for reporting this issue! I will have a look at it tomorrow (or next week).

Thanks,
Andre

opc241:
Hi Andre, 

i could absolutely confirm your idea and tested it by deleting two .nfo files and after that the scraper displays those two movie files. 
I had a look to the .nfo files and there were no <media> tags present (as mentioned by the log file). 
But newertheless, especially in this case i would need MediaElch :)

bugwelle:
Thanks for confirming it. I'll prepare a fix and let you know when a new version is available. :)

bugwelle:
Reproduced locally:

- scan movies
- check movie X exists
- go to movie X's directory
- `touch movie.nfo`
- reload
- confirm movie is "gone" (it is listed, but without any title)

[...]

PR #1567, opened 2023-04-30
fix(file searcher): Don't ignore movies if their NFO can't be parsed

bugwelle:
For #1564


bugwelle:
This commit broke image loading from disk. See https://github.com/Komet/MediaElch/issues/1572

PR #1570, opened 2023-05-04
fix(UI): Selection background color of shows/episodes and artists/albums

### LINKED ISSUE #1569: Selected item on TV Show non-visible
**Describe the bug**
When an item is selected,  colors are not set properly, thus becoming non-visible (either the item or the selection).

**To Reproduce**
1. Go to TV Shows
2. Click on any item (Show, season or episode).

**Expected behavior**
Follow the same behavior than other items at other tabs.

**Screenshots**
Light UI, item becomes invisible
image

Dark UI, no selection is shown
image

**MediaElch Version:**
 - [x] 2.10.0 stable

**Operating System:**
 - [x] Windows
 - [ ] macOS

**Additional context**
Windows 7

bugwelle:
It seems due to strange behavior of Qt, settings the background color for selections via `selection-background-color` only affects `PE_PanelItemViewRow` on macOS, whereas `PE_PanelItemViewItem` is affected by `background-color` and `::item:selected` on Windows/Linux.

See https://stackoverflow.com/a/34646515/1603627

This is strange. I don't get why this is the case, yet.

Fix #1569

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
  - UI:
+   - Selecting TV shows / episodes or artist / albums now has proper background colors on Windows and macOS (#1569) 
    - Color labels have better color for the dark theme (#1545)
    - The language dropdown menu is now sorted according to the translated language names (#1560)
    - Fix the ordering of custom movie scraper details (previously sorted randomly)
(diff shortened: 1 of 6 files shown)


bugwelle:
Result on Linux (Kubuntu 22.04):

Screenshot_20230504_204336

Issue #1575, opened 2023-05-25
genres placeholder for renamer


schoenid:
**Is your feature request related to a problem? Please describe.**
I've used tinyMediaManager so far. My movie library is organized like `${genres[0]}/${movieSet.title}/${title} (${year})` in terms of tinyMediaManager. 
I've seen that it's not possible to do the same in MediaElch, because there is no **genres** placeholder for the renamer.

**Describe the solution you'd like**
It would be nice to have a conditional placeholder like `{genres}<genres[0]>{/genres}` for MediaElchs renamer.
I see only one problem in the way, like MediaElch manages genres like tags. There would also a possibility be required, to allow to sort these genre tags, to be able to address one as the main genre. To select a genre tag with "Ctrl"-Click would be a possibility, to make it the first ...

**Describe alternatives you've considered**
Alternatively I have either to manage the library manually by moving files and folders in a file manager, or to continue to work with tinyMediaManager.

schoenid:
Hi @bugwelle 

> Do you know if TinyMediaManager has conditionals such as` {dvd}DVD{/dvd}`?

Not in version 3.
(I don't  want to upgrade ...)

Thanks for your answer.
Best Regards, Daniel

bugwelle:
Thanks.

I write about a new concept here a bit: https://github.com/Komet/MediaElch/pull/1576
Feel free to add comments. It's work in progress.

Regards,
Andre

PR #1576, opened 2023-05-27
doc(Concepts): Write about a new renamer concept

bugwelle:
Work in progress

### DIFF
--- a/docs/concepts/renamer.md
+++ b/docs/concepts/renamer.md
@@ -1,0 +1,268 @@
+# Renamer
+
+__Status__: Concept, work in progress  
+__Last Updated on__: 2023-07-02
+
+New features for our movie/TV show/… renamer were requested on a regular basis.
+Over the last few years, those requests have culminated.  This document tries
+to propose a new renamer, that is more dynamic than the old one, which required
+hard-coded placeholders for both conditions and variables.
+
+
+## Current State
+
+The following paragraphs describe the current state, including some drawbacks.
+
+### Error Handling
+
+There exists only minimal error handling and no proper heuristics whether
+a proposed file name actually makes sense.  For example, renaming a movie that
+hasn't been scraped, yet, could lead to a filename that is just ` ().mp4`,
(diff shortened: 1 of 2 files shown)


docs/concepts/renamer.md
```diff
@@ -0,0 +1,251 @@
...
+character, one could use `<actors[0].role>`.
+
+In this example, `actors` is a list (see "Index Operator"), and its elements
+are "actor" _structures_.
+
+The `.` allows access into a structure's fields.
+
+Imagine that a movie had a structured field `rating` that is not a list.
+If we want to access the count of votes, we could use `rating.voteCount`.
+
```
silvus, line 219: How can you handle genres ?

For exemples with IMDB scrapper :
- "Gladiator" will have genres "Action, Adventure, Drama", sort by alphabetical order. So `<genres[0]>` will give a "Action" result.
- "Black Hawk Down" will have genres "Action, Drama, History", sort by alphabetical order. So `<genres[0]>` will give a "Action" result.
- "Pacific Rim" will have genres "Action, Adventure, Sci-Fi", sort by alphabetical order. So `<genres[0]>` will give a "Action" result.
- "Die Hard"  will have genres "Action, Thriller", sort by alphabetical order. So `<genres[0]>` will give a "Action" result.

Each of these films would be better defined by any other genre than the very generic "Action".
Perhaps it would be interesting to allow genres to be reordered by film to put the most relevant first?
It will be the same on the tags.

bugwelle: Hi, sorting is a requirement for this to work. It was suggested in https://github.com/Komet/MediaElch/issues/1575 :)
Since Kodi's NFO file's use simple XML tags for genres, we can sort them internally according to the NFO file. A next step would be to allow sorting genres via the UI.

I have added a sentence or two about that to this document.

bugwelle:
Merging this document. It's not been implemented, yet, so new changes will probably come in the future.

PR #1603, opened 2023-07-08
fix(MusicBrainz): Don't set Content-Type header in GET request

### LINKED ISSUE #1597: MediaElch don't scrapping
**Scrapers that don't work:**

-[]Universal Music Scraper

**Details which are wrong / not loaded:**
Add a list of information that is not loaded correctly.
in the window of the scrapping do not appear anyone artist 

**MediaElch Version:**

 - [ ] 2.10.2 stable

**Operating System:**
 - [ ] Windows

**Additional context:**
I already was having problems with the details of discography (don Loaded), and today knew about latest version. I installed with chocolatey an instance of MediaElch with console (admin) without uninstalling the previous one, but it only upgrade MediaElch to 2.10.0 version. Then I deleted the previous one manually and uninstalling the chocolatey version by console, all to upgrade with the newest version of Github. I Copy the folder in programfiles folder, and it
Bartok-don't-appears
 appears with the problem.

bugwelle:
It appears as if MusicBrainz can't handle search requests with a Content-Type header anymore, at least not XML.

Without the header, requests work as expected.

Fix #1597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - Movie scraper: Original title was missing and the setting "ignore duplicate original title"
   was ignored (#1601).
+- Universal Music Scraper: The MusicBrainz part of the universal music scraper now works again (#1597).
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


Santicoma:
I Understand, but, what have I to do?

bugwelle:
@Santicoma Should have mentioned that you can / have to download the latest snapshot version. The next MediaElch release will also contain this fix.

See https://mediaelch-downloads.ameyering.de/snapshots/ :)

Issue #1640, opened 2023-08-29
Slow UI update when changing movies


fungos:
**Describe the bug**
Changing movie selection for movies that did not load from database yet, is very slow.

**To Reproduce**
1. Go to 'Movies' screen
2. Click on any movie that is not on DB (has a Star before its name)
3. Click on any other movie that is also not on DB or loaded before
4. See how long it takes to change from one movie to another

**Expected behavior**
Fast responsive UI

**MediaElch Version:**
 - [X] 2.10.5-dev (nightly)
 - [ ] 2.10.4 stable

**Operating System:**
 - [ ] Windows
 - [ ] macOS
 - [X] Linux (distribution: Manjaro)

**Additional context**
![image](https://github.com/Komet/MediaElch/assets/151544/597c9ec6-0af3-4272-9438-6f911e90c537)
![image](https://github.com/Komet/MediaElch/assets/151544/afd1d689-7033-466a-bb57-f8a330c0ee8b)

Doing this change improves a little, still not fast enough. The issue is the amount of widgets that will receive and process the event `dataChanged`

bugwelle:
This has been in MediaElch right from the beginning and has annoyed me quite a bit :smile:

The amount of "sigChanged" signals is _a lot_ when a movie is selected. Adding a log entry:

```
2023-09-07 11:37:13.070 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: false
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
2023-09-07 11:37:13.084 DEBUG [../MediaElch/src/data/movie/Movie.cpp:791] | sigChanged: true
[...]

PR #1643, opened 2023-09-07
feat: Load images asynchronously in the UI

bugwelle:
For #1640

------

Load images in `ClosableImage` asynchronously if a file from disk is requested.

--------

@fungos  You can try out this PR to see whether performance improves on your side as well. In my case, performance has been drastically increased for movies that have huge posters/banners.

Note, however, that I didn't implement it _properly_ yet. This is for testing only as of now. :)
There may be race conditions if movies are switched too fast and I've practically disabled the image cache. That will come later. It's a proof-of-concept for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     For example, the country's release date will be used.
   - If a movie has more than one certification in a given language, we sometimes accidentally
     ignored the selected language if TMDB has an empty certification for it (#1641)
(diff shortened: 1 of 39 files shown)


fungos:
Great work, this fixed the issue. There is no more delays switching movies.

bugwelle:
- [ ] TODO: Actually cache the image; needs some refactoring

PR #1653, opened 2023-10-02
Fix #1547 Exclusion filter does not exclude subfolder

### LINKED ISSUE #1547: Exclusion filter does not exclude subfolder
**Describe the bug**
Folder excluding filter in advancedsettings.xml doen't exclude subsfolders.

**To Reproduce**
1. Define a folder excluding pattern in advancedsettings.xml.
2. Eg. `<pattern applyTo="folders">[.]ignore$</pattern>`
3. Start MediaElch.
4. Perform rescan.

**Expected behavior**

Expectation: The whole Movie's path's directory, including its subfolders, is filtered and excluded.
Actual: Only direct parent (just files in upper level) directory is filtered.

For example, when the pattern is <pattern applyTo="folders">[.]ignore$</pattern>,

"/user/My Videos/.ignore/test.vob" is excluded, but,
"/user/My Videos/.ignore/test3/video1/test-003.vob" is not excluded.

**MediaElch Version:**
 -  2.10.0

**Operating System:**
Windows 10

**Additional context**

I found MediaElch filter exclusion is somewhat different from Emby.  In emby, if there is a folder renamed to exact match of ".ignore" ,  the entire folder and its subfolders will be excluded from scanning.   Emby even allows file name detection, which means that if you rename any file (note: not folder) to ".ignore" in say "Folder A", anything inside "Folder A" (including things in subfolders) will be excluded from scanning. Is this achievable in MediaElch?  Even if it's not exactly like that, I suppose one would expect that subfolders exclusion should work properly?

Neinei0k:
I already described the reason for this problem in the issue comment, but I copy it here for completeness.

I tracked this problem to [MovieDiskLoader::loadMovieContents()](https://github.com/Komet/MediaElch/blob/f12f823075685d9e247ae9ae16a17a86783efa60/src/file_search/movie/MovieDirectorySearcher.cpp#L132). It uses QDirIterator to generate a list of files and directories, each of them is checked against exclusion filters and a bunch of other hardcoded filters. But all of them only check the file fileName or its direct parent dirName. Parents of parent directory are not checked.

I've tried to implement a solution provided by @bugwelle. Instead of using a single QDirIterator which also iterates through sub-directories, a QDirIterator is created for each directory. Filters are applied to files and directories and whichever directory passes, will be added into the queue to be checked next.

There are a few minor optimizations that could have being done but I decided to mostly leave everything as is, so not to introduce additional bugs. I've made one change - combined hardcoded folder exclusion filters and I added "featurettes" folder to also be excluded, since it appears with movie files quite often.

I've tested this against my own library and it seems to work fine.

### DIFF
--- a/src/file_search/movie/MovieDirectorySearcher.cpp
+++ b/src/file_search/movie/MovieDirectorySearcher.cpp
@@ -131,106 +131,104 @@ bool MovieDiskLoader::doKill()
 
 void MovieDiskLoader::loadMovieContents()
 {
-    QDirIterator it(m_dir.path.path(),
-        m_filter.fileGlob,
-        QDir::NoDotAndDotDot | QDir::Dirs | QDir::Files,
-        QDirIterator::Subdirectories | QDirIterator::FollowSymlinks);
-
+    QQueue<QString> dirs;
+    dirs.enqueue(m_dir.path.path());
+    
     QString lastDir;
-
-    while (it.hasNext()) {
-        if (isAborted()) {
-            return;
-        }
-        it.next();
-
-        QString dirName = it.fileInfo().dir().dirName();
-        QString fileName = it.fileName(); // may actually be a directory name
-
-        const bool isFile = it.fileInfo().isFile();
-        const bool isDir = it.fileInfo().isDir();
-        bool isSpecialDir = false; // set to true for DVD or BluRay Structure
-
-        if (isFile && m_filter.isFileExcluded(fileName)) {
-            continue;
-        }
-
-        // TODO: If there is a BluRay structure then the directory filter may not work
-        // because BDMV's parent directory is not listed.
-        if ((isDir && m_filter.isFolderExcluded(fileName)) || m_filter.isFolderExcluded(dirName)) {
-            continue;
-        }
-
-        // Skips Extras files
-        if (isFile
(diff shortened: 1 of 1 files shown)


bugwelle (approved):
Thank you very much! The change was a lot smaller than I expected. Tested it locally and works just fine.

I will merge this and update the changelog in a subsequent commit.

bugwelle:
@Neinei0k I would mention you in the changelog if that is fine with you. Something like:

> ### Fixed
> …
>  - movies: Exclusion filters are now applied to all folders, not just a file's parent directory (#1653)  
>    Thanks to GitHub user `Neinei0k` for implementing it!

src/file_search/movie/MovieDirectorySearcher.cpp
```diff
@@ -131,106 +131,104 @@ bool MovieDiskLoader::doKill()
...
+                continue;
+            }
 
-        // Skip extra thumbs folder and all files inside it
-        if (QString::compare("extrathumbs", dirName, Qt::CaseInsensitive) == 0) {
-            continue;
-        }
+            // Skip folders and all files inside them
+            if (isDir
+                && (QString::compare(".actors", fileName, Qt::CaseInsensitive) == 0
```
bugwelle, line 184: Btw, good thing to group it. The comments on the left side were unnecessary.

PR #1655, opened 2023-10-04
Fix #1635 Duplicate movie Genres, Countries and Studio labels when scrapping

### LINKED ISSUE #1635: Duplicate movie Genres, Countries and Studio labels when scrapping
**Describe the bug**
Movie Scrapper concatenates/append Genres, Countries and Studio labels instead of replacing them.

**To Reproduce**
Tested on Windows 11. Tested on a fresh Ubuntu 2023.04 as well and all default settings.
1. Select a movie from the list
2. Move on to the Extended tab. Note the Genre, Countries and Studio tags
3. Scrap the infos. Wait for the display to be updated.
4. If the movie details are new, repeat previous step.
5. See that the new details are added to the previous, instead of being replaced.
6. Save the movie details
7. The XML file also contains the duplicated entries.

**Expected behavior**
The scraper should replace the values... or append new labels only.

**Screenshots**
N/A

**MediaElch Version:**
 - [ ] 2.10.5-dev (nightly)
 - [X] 2.10.4 stable

**Operating System:**
 - [X] Windows 11
 - [ ] macOS
 - [X] Linux (distribution: Ubuntu 2023.04)

**Additional context**
When clicking on one of the duplicate tag, it is removed (and the others are well kept), as expected... However the display does not correspond to the reality. Save the movie and see that the removed tag - as well as its duplicates entries - are removed.

Hope this helps, let me know if further details are needed.

Neinei0k:
Tags are set the same way as genres, countries and studios, but didn't have duplication problem because there was a check if the tag is already set for the movie. I added the same check to genre, country and studio. Probably not necessary, but I also added test for empty argument into `Movie::addTag`, since it was done in the other functions.

It probably makes more sense to use `QMap` instead of `QStringList` for tags, genres, countries, and studios. It would make checking for existing items faster, and when iterating `QMap`, the items will be sorted by keys. But I'm not sure how much it would affect the performance, and it would probably require a lot of changes.

### DIFF
--- a/src/data/movie/Movie.cpp
+++ b/src/data/movie/Movie.cpp
@@ -852,7 +852,7 @@ void Movie::addActor(Actor actor)
  */
 void Movie::addCountry(QString country) // NOLINT // we require pass-by-value
 {
-    if (country.isEmpty()) {
+    if (country.isEmpty() || m_countries.contains(country)) {
         return;
     }
     m_countries.append(country);
@@ -866,7 +866,7 @@ void Movie::addCountry(QString country) // NOLINT // we require pass-by-value
  */
 void Movie::addGenre(QString genre) // NOLINT // we require pass-by-value
 {
-    if (genre.isEmpty()) {
+    if (genre.isEmpty() || m_genres.contains(genre)) {
         return;
     }
     m_genres.append(genre);
@@ -880,7 +880,7 @@ void Movie::addGenre(QString genre) // NOLINT // we require pass-by-value
  */
 void Movie::addStudio(QString studio) // NOLINT // we require pass-by-value
 {
-    if (studio.isEmpty()) {
+    if (studio.isEmpty() || m_studios.contains(studio)) {
         return;
     }
     m_studios.append(studio);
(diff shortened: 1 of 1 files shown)


bugwelle (approved):
LGTM. Thanks. :)

Though this may have side effects such as trying to add the same tag twice in the UI -> it would just disappear. But a good workaround for now. :)

> It probably makes more sense to use `QMap` instead of `QStringList` for tags, genres, countries, and studios. It would make checking for existing items faster, and when iterating `QMap`, the items will be sorted by keys. But I'm not sure how much it would affect the performance, and it would probably require a lot of changes.

While `QMap` would make sense, we would loose the order of tags/genres/etc.
The order is specified as the order of the corresponding XML tag in NFO files.  This isn't important _now_, but some users requested to be able to sort genres on a per-movie-basis, which would then become important for the movie renamer, where I want to add something like a `genres[0]` placeholder.

Neinei0k:
> Though this may have side effects such as trying to add the same tag twice in the UI -> it would just disappear. But a good workaround for now.

I think it worked like this before. There is the same check for duplicate tags in the [TagCloud widget](https://github.com/Komet/MediaElch/blob/65085b373378ed870dbb9ee85922868e9e7a2362/src/ui/small_widgets/TagCloud.cpp#L185).

> While `QMap` would make sense, we would loose the order of tags/genres/etc. The order is specified as the order of the corresponding XML tag in NFO files. This isn't important _now_, but some users requested to be able to sort genres on a per-movie-basis, which would then become important for the movie renamer, where I want to add something like a `genres[0]` placeholder.

That makes sense. They are shown sorted, so I assumed that they are also saved sorted in .nfo file.

PR #1697, opened 2023-12-27
chore(deb): Update debianization

iluminat23:
- Drop CDBS and debhelper with cmake instead
- cleanup dependencies
- Use external libquazip1-qt5 (requieres Debian >= Bookworm || Ubuntu >= Lunar)


debian/control
```diff
@@ -3,11 +3,10 @@ Section: misc
 Priority: optional
 Maintainer: Andre Meyering <info@andremeyering.de>
 Standards-Version: 4.6.2
-Build-Depends: debhelper (>= 13.0.0),
-               cdbs,
+Build-Depends: debhelper-compat (= 13),
```
bugwelle, line 6: Thanks. I had issues with debhelper-compat locally, so switched back to debhelper + compat file. 👍

debian/control
```diff
@@ -19,31 +18,16 @@ Build-Depends: debhelper (>= 13.0.0),
...
-         libqt5network5,
-         libqt5opengl5,
-         libqt5sql5,
-         libqt5sql5-sqlite,
-         libqt5svg5,
-         libqt5xml5,
-         libzen0 | libzen0v5,
-         zlib1g,
+Depends: ${misc:Depends},
+         ${shlibs:Depends},
```
bugwelle, line 30: Could you bring me up to speed here? Why are the runtime dependencies not listed here? At least libmediainfo is loaded at runtime and auto-detection (if any used) fails. That's at least what's happening with openSUSE: https://github.com/Komet/MediaElch/blob/ad330574c22a421524312655b2c52acf86f6509e/obs/MediaElch.spec#L12

debian/rules
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/make -f
 
-QT_SELECT=qt5
-export QT_SELECT
+export DEB_BUILD_MAINT_OPTIONS = hardening=+all
```
bugwelle, line 3: I thought those flags would be passed to the build via environment variables. Isn't it best practice to rely on env variables passed by e.g. launchpad.net's build system instead of hard-coding them ourselves?

(I've never fully read the debian maintainer guide, so no idea what's best practice :) )

debian/rules
```diff
@@ -1,8 +1,14 @@
...
+export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
+export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
-include /usr/share/cdbs/1/rules/debhelper.mk
-include /usr/share/cdbs/1/class/qmake.mk
+export QT_SELECT = 5
```
bugwelle, line 7: Isn't this bash-like? Are whitespace characters really allowed after and before `=`? (they're not in bash)

bugwelle (commented):
Thank you very much for this PR! You're the hero I need! 😄 

I had to revert to the [previous `debian/` files](https://github.com/Komet/MediaElch/commit/647ad7ef7d1cc96a48a5af9ac556178ebdc641e2) while packaging the latest version, because a few things didn't work properly (I _think_ because I still wanted to build for Ubuntu 18/20).
I still have a few dozen Wiki pages open in my browser: all about Debian/Ubuntu maintainer guides / packaging guides that I never good to ready 100%.

So I very much appreciate your PR! Before approving (and merging), I still have a few questions. It would be awesome if you could answer them, especially because it helps me to understand the packaging process better.

Thanks!

debian/control
iluminat23: If you still want to build for Ubuntu 18, we need to go to debhelper 11. If i recall this correctly it is since debhelper 9 that the compat file can be dropped and `debhelper-compat` can be used as dependency.

debian/rules
iluminat23: The package maintainer should know best which flags are safe to use for the package. So the build environment doesn't impose much on the package.
Make doesn't mind the spaces and this is taken directly from the [Guide for Debian Maintainers](https://www.debian.org/doc/manuals/debmake-doc/ch08.en.html#cmake-single)
The _Debian New Maintainers' Guide_ is replaced by the _Guide for Debian Maintainers_. The new guide is much better, and has more and better examples.

iluminat23:
As you wrote you need to also package for Ubuntu 18 and 20. You bundle the `libquazip1-qt5`. And this is a bit tricky. For system which have a `libquazip1-qt5` this leads to collisions. The debianization for systems which have and those without `libquazip1-qt5` needs to differ.

The debianization I did is not working with the bundled `libquazip1-qt5`. Give me a bit to find a solution for this.

iluminat23:
A question out of the direct scope of this ticket. Is libcurl anywhere used? I greped through the git repo and didn't find any reference that libcurl is used. Only the linker flags. Or is it implicit through some dependency?

debian/control
bugwelle: No, Ubuntu >= 20 is fine. Builds failed for Ubuntu 18 anyway, because of some missing header file (some header has incorrect casing, e.g. `Qmultimedia/` instead of `QMultiMedia/` and I never found the root cause). Also, IIRC, Ubuntu 18 is EOL. If there are still users on Ubuntu 18, they can use the AppImage. :)

debian/control
bugwelle: Oh my god, that is awesome. I wasn't aware of that! I read that page once, but apparently didn't read it properly 😄

debian/rules
bugwelle: Great, thanks! All those flags are what I expected launchpad.net to add anyway.

I think we could even switch to Qt 6. I will use the "old" `debian/` files for Ubuntu 20 and later anyway.
But you don't need to do that! I want to update all build-instructions to Qt6 next year anyway and will try to update this then.

I also thought about moving these files to a separate "packaging" repository once. But I'm still unsure whether that's a good idea. 😄

bugwelle:
>  Or is it implicit through some dependency?

IIRC, only implicit through Qt (unchecked, though!). We only use Qt's network capabilities, no raw curl commands. I'm not the original author of the debian packaging files, so no idea why it was added :)

Thank you very much for your detailed responses. I'm currently visiting my family, so can't test it myself (using the https://launchpad.net/~mediaelch/+archive/ubuntu/mediaelch-test repo), but will do so in January. :)

debian/rules
iluminat23: In my company we use http://honk.sigxcpu.org/projects/git-buildpackage/manual-html/gbp.html and different branches for the different distributions.
Then the debianization is located only on the distribution branch. This mirrors the way Debian/Ubuntu are managing packages. As they use the software from upstream and put a debianization on top of it.

iluminat23:
Hi, sorry for the delay.
I updated the debianization:
- drop libcurl (also from `MediaElch.pro` and `src/CMakeLists.txt`
- drop libpulse from build deps
- readded some build/runtime dependencies
- added `libxkbcommon-dev` to the dependencies
- use Qt6 now

This works for Debian Bookworm ans should work for Ubuntu >= 22

bugwelle:
> Hi, sorry for the delay.

Don't be! I still on vacation and didn't get a proper look at this PR, yet. My TODOs from https://github.com/Komet/MediaElch/pull/1697#issuecomment-1870098647 are still open.

> This works for Debian Bookworm ans should work for Ubuntu >= 22

You're awesome, thanks!

I'll review it properly on the weekend/ next week. But so far (on a quick glance), it looks good to me.

Would you be ok with me mentioning you in the changelog with a small note saying something along the lines of "updated debian packaging to Qt6, thanks to GitHub user `iluminat23`"? (or your real name or no mention at all; up to you)

Regards,
Andre

iluminat23:
> I'll review it properly on the weekend/ next week. But so far (on a quick glance), it looks good to me.

I just wanted to do this before I'm back at work. Take your time!

> Would you be ok with me mentioning you in the changelog with a small note saying something along the lines of "updated debian packaging to Qt6, thanks to GitHub user iluminat23"?

I'm fine with either.

Best regards,
Philipp

bugwelle (approved):
Thank you very much! It works with Ubuntu 23.04 (Lunar).
I will have a look at your links later (I'll create a backlog item for that)


Notes

- [x] release build works
  - [x] cmake
  - [x] qmake
- [x] debug build works
  - [x] cmake
  - [x] qmake
- [x] local packaging works (` ./.ci/linux/package_linux_launchpad.sh deb`)
  - Note: I had to use a Docker container. I'm still on Kubuntu 22.04, which does not have some qt6 packages
  - I had to adapt the packaging script; nothing for you to worry about; I'll update it in a separate PR :) (see https://github.com/iluminat23/MediaElch/pull/1 which I accidentally opened)
- [x] launchpad packaging works (` ./.ci/linux/package_linux_launchpad.sh launchpad`)
  - [x] can be uploaded
  - [x] can be build on launchpad.net
  - [x] can be installed via apt

Other Notes

<details>
```
W: mediaelch source: inconsistent-appstream-metadata-license data/desktop/com.kvibes.MediaElch.metainfo.xml (cc-by-4.0 != lgpl-3.0-only) [debian/copyright]
W: mediaelch source: no-debian-changes
```

Build via https://launchpad.net/~mediaelch/+archive/ubuntu/mediaelch-test/+build/27615875
```
Accepted:
 OK: mediaelch_2.10.7-dev.12.orig.tar.xz
 OK: mediaelch_2.10.7-dev.12-1~lunar.debian.tar.xz
 OK: mediaelch_2.10.7-dev.12-1~lunar.dsc
     -> Component: main Section: misc

mediaelch (2.10.7-dev.12-1~lunar) lunar; urgency=medium

  * next build
```

```
$ apt search MediaElch
Sorting... Done
Full Text Search... Done
mediaelch/lunar,now 2.10.7-dev.12-1~lunar amd64 [installed]
  Media Manager for Kodi
```

</details>

bugwelle:
Changelog entry via https://github.com/Komet/MediaElch/commit/b3d7b7c608dcd7cd553b656a8fe34096136257aa

PR #1733, opened 2024-04-06
fix(fernsehserien.de): Fix scraping of episode thumbnails

bugwelle:


test/resources/scrapers/tmdbtv/The-Simpsons-S12.ref.txt
```diff
@@ -1798,7 +1798,7 @@ actors: (N>5)
  - id: 6008
    name: Harry Shearer
```
bugwelle, line 1799: Harry Sharer no longer has an image on TMDB

PR #1760, opened 2024-06-22
feat(Concerts): Support multiple `<artist>` tags

### LINKED ISSUE #1709: Music videos: Support for multiple `<artist>` tags
### Discussed in https://github.com/Komet/MediaElch/discussions/1706

<div type='discussions-op-text'>

<sup>Originally posted by **rainerhillebrand** January 20, 2024</sup>
The table in the article https://kodi.wiki/view/NFO_files/Music_videos in section "Music Video Tags" defines that multiple <artist> tags are supported in nfo files for Music Videos.
MediaElch seems to support only a single <artist> tag so that multiple artists can only be given in a single string.
When showing all artists of all Music Videos in Kodi (at least in Nexus 20.3) then you cannot select one of these multiple artists. Additional information about each artist cannot be shown via the music video item's context menu.
Wouldn't it make sense that MediaElch supports multiple <artist> tags for Music Videos nfo files as well?
</div>

bugwelle:
Fix  #1709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@
 - Movies have a new field "TV Show Link" which is stored as a `<showlink>` in Kodi.  
   It is available in the UI via the movie's "Extended" tab (#1753).
 - Added Kodi v22 to Kodi selection menu.
+- Concerts (i.e. music videos in Kodi) can now have multiple artists, each stored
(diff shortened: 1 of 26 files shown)


rainerhillebrand:
Thanks a lot!

Issue #1082, opened 2020-12-14
Renamer: Allow replacing spaces with other delimiters


bugwelle:
**Is your feature request related to a problem? Please describe.**
Spaces in filenames are sometimes not wanted. It should be possible to replace them with other delimiters.

**Describe the solution you'd like**
I want to be able to replace spaces with `_`, `.` and other characters.

**Additional context**
Requested on the Kodi Forum. See https://forum.kodi.tv/showthread.php?tid=136333&pid=2995289#pid2995289

norg:
I would also like this feature :) I migrated my DB from TMM to MediaElch now and quite happy. Currently I run a bash script to convert the spaces to underscore.

Can you give some hints for someone new to the project but willing to try to implement it?

bugwelle:
Hi,

thank you very much for your interest in MediaElch. :-)

There are few files that are of interest. `RenamerDialog.cpp` (+`.h` and `.ui`) as well as `Renamer.cpp` and its variants (`MovieRenamer.cpp`, `ConcertRenamer.cpp` etc) in `src/renamer`).

You'd need to change the UI to add a way to configure space->char transformation as well as need to modify the appropriate classes (`MovieRenamer`, etc).

For building MediaElch, see https://mediaelch.github.io/mediaelch-doc/contributing/build/index.html

Regards,
Andre

taniashafiei:
Hello, I also found this feature interesting and I implemented it in the PR #1827 :D .

bugwelle:
Implemented by @taniashafiei 

The next snapshot version will contain this feature :)

Issue #1710, opened 2024-01-24
Looking for new contributors/maintainers


bugwelle:
Next month, I've been working on MediaElch for about 6 years. Many new features, bug fixes and other new stuff was implemented in that time. We've migrated to new Kodi versions, new Qt (UI framework) versions, and kept MediaElch working.

However, you may have noticed that more issues have been opened over the last few months than have been fixed. I don't have the time anymore to fix all issues, respond to comments in a timely manner and to develop new features.

__I'm looking for new contributors and for new (co-)maintainers.__

If you're interested in C++ and Qt, MediaElch may be the place to work on.
I'll help you get started, will review code, etc.

I'm still working on MediaElch, but will limit my contributions to bug fixes such as scraper-fixes and other small issues. New features are not on my roadmap anymore and I'll refer feature requests to this issue.

MediaElch has been and still is a fun project. I love the community: people have always been nice and great to work with. But as always, life goes on: new projects, private life, work life, and more change.

~ Andre

bugwelle:
Hi all,

as my first child was born last month, I didn't have the time to look at MediaElch at all until now.
I see that there are many issues with scrapers, but sadly I don't have any free time in the foreseeable future.

Kind regards,

Andre

PR #1827, opened 2024-11-29
feat: replace spaces by another delimiter in renamer

taniashafiei:
add feature from request #1082

### DIFF
--- a/src/renamer/Renamer.cpp
+++ b/src/renamer/Renamer.cpp
@@ -71,6 +71,15 @@ QString Renamer::replaceCondition(QString& text, const QString& condition, bool 
     return text;
 }
 
+bool Renamer::replaceDelimiter(QString& text, QString& oldDelimiter, QString& newDelimiter, bool execCondition)
+{
+    if (execCondition == true){
+        text.replace(oldDelimiter, newDelimiter);
+        return true;
+    }
+    return false;
+}
+
 bool Renamer::rename(const QString& file, const QString& newName)
 {
     QFile f(file);
--- a/src/renamer/Renamer.cpp
+++ b/src/renamer/Renamer.cpp
@@ -71,7 +71,7 @@ QString Renamer::replaceCondition(QString& text, const QString& condition, bool 
     return text;
 }
 
-bool Renamer::replaceDelimiter(QString& text, QString& oldDelimiter, QString& newDelimiter, bool execCondition)
+bool Renamer::replaceDelimiter(QString& text, QString oldDelimiter, QString newDelimiter, bool execCondition)
 {
     if (execCondition == true){
         text.replace(oldDelimiter, newDelimiter);
--- a/src/settings/Settings.cpp
+++ b/src/settings/Settings.cpp
@@ -817,6 +817,18 @@ void Settings::renamePatterns(RenameType renameType,
     seasonPattern =
         settings()->value(QString("RenamePattern/%1/SeasonPattern").arg(renameTypeStr), "Season <season>").toString();
 }
+void Settings::renamePatterns(RenameType renameType,
+    QString& fileNamePattern,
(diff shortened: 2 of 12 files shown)


bugwelle:
Hi!

First of all: THANK YOU for opening a PR! That is very much welcome and I'm glad other people are interested in developing MediaElch!

Due to the size of this PR, I will likely not get to review this this week. See https://github.com/Komet/MediaElch/issues/1710 for reasons (time constraints).

The renamer is _old_ and not well written. The fact that you had to adapt so many places is a sign for that. :)

Do you want me to change minor stuff like renaming `defaultValue` to `defaultDelimiter` myself or should I add code comments, so that you can pick it up? Whatever suites you best.

Regards,
Andre

src/renamer/Renamer.cpp
```diff
@@ -71,6 +71,15 @@ QString Renamer::replaceCondition(QString& text, const QString& condition, bool
     return text;
 }
 
+bool Renamer::replaceDelimiter(QString& text, QString oldDelimiter, QString newDelimiter, bool execCondition)
```
bugwelle, line 74: The third argument seems strange. I think it would be better to remove it and move the boolean into a condition surrounding the call to `replaceDelimiter`:

```cpp
if (execCondition) {
  Renamer::replaceDelimiter(...)
}

Reason being: It is not clear to the caller what the third argument does.

Also: `sanitizeFolderName` already calls `sanitizeFileName` in your PR. Why call `sanitizeFileName` after `replaceDelimiter` in `*Renamer.cpp`? :)

File: src/settings/Settings.cpp [hunk 2]
```diff
@@ -847,6 +876,13 @@ void Settings::renamings(RenameType renameType, bool& files, bool& folders, bool
     seasonDirectories =
         settings()->value(QString("RenamePattern/%1/UseSeasonDirectories").arg(renameTypeStr), true).toBool();
 }
+void Settings::renamings(RenameType renameType, bool& files, bool& folders, bool& seasonDirectories, bool& replaceDelimiter)
```
bugwelle: There is no need to add another method. Adapting the existing one is fine. :)
The code isn't great to begin with, so adding another argument is ok here.

src/settings/Settings.cpp
```diff
@@ -838,6 +861,12 @@ void Settings::setRenamings(RenameType renameType, bool files, bool folders, boo
     settings()->setValue(QString("RenamePattern/%1/RenameFolders").arg(renameTypeStr), folders);
     settings()->setValue(QString("RenamePattern/%1/UseSeasonDirectories").arg(renameTypeStr), seasonDirectories);
 }
+void Settings::setRenamings(RenameType renameType, bool files, bool folders, bool seasonDirectories, bool replaceDelimiter)
+{
+    Settings::setRenamings(renameType, files, folders, seasonDirectories);
+    const QString renameTypeStr = renamerTypeToString(renameType);
+    settings()->setValue(QString("RenamePattern/%1/ReplaceDelimiter").arg(renameTypeStr), replaceDelimiter);
```
bugwelle, line 848: I'd prefer `Delimiter`, because that's what it is intended to be: a delimiter between words. That we only replace spaces is a restriction or implementation detail.

bugwelle (approved):
Hi,

thanks! I've pushed a commit that reviews/changes this PR a bit. I'll leave comments as well, but note that those are fixed in my commit and are only meant for reference. :)

Regards,
Andre

src/settings/Settings.cpp
bugwelle: Ah, this comment was meant for the string value, not the boolean value. :)

taniashafiei:
Thank you for your feedback, the correcting and accepting the pull request !

Issue #1842, opened 2025-01-12
Loosen 2k Detection to enable detection of letterbox


sharkbit:
Would like to open up the 2160 resolution detection to support detection Leterbox

https://github.com/Komet/MediaElch/blob/5bff30e88f012ea12003681402246dd19e34825f/src/globals/Helper.cpp#L496

I would suggest 
  `  } else if (height >= 1570 || width >= 3820) {`

refrence: [aspect-ratio-cheat-sheet](https://www.unravel.com.au/aspect-ratio-cheat-sheet)

bugwelle:
Fixed via #1843

PR #1843, opened 2025-01-12
feat: Loosen 2160p detection to support letterbox

sharkbit:
for #1842

### DIFF
--- a/src/globals/Helper.cpp
+++ b/src/globals/Helper.cpp
@@ -493,7 +493,7 @@ QString matchResolution(int width, int height, const QString& scanType)
     QString res;
     if (height >= 4312 || width >= 7672) {
         res = "4320";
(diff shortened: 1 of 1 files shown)


src/globals/Helper.cpp
```diff
@@ -493,7 +493,7 @@ QString matchResolution(int width, int height, const QString& scanType)
     QString res;
     if (height >= 4312 || width >= 7672) {
         res = "4320";
-    } else if (height >= 2152 || width >= 3832) {
+    } else if (height >= 1570 || width >= 3820) {
```
bugwelle, line 496: Ah, because of ?`3840 x 1572` of your linked table? :)

bugwelle (approved):
LGTM, thanks. :)

Issue #1859, opened 2025-02-14
Top menu icons don't show: libqt6svg6 missing as dependency in apt repository


dmackca:
**Describe the bug**
I installed MediaElch this evening from the apt repository onto my media server (Linux Mint XFCE 22.1)
I access it by SSH with X forwarding (`ssh -X <my server>`) or via RDP through Remmina. 
The icons on the top menu don't show up. The icons in the left sidebar show just fine, though.

**To Reproduce**
1. `ssh -X <my server>`
2. `MediaElch`
3. Observe that the icons in the top menu are invisible.

**Expected behavior**
I've used MediaElch in ZorinOS and Windows 11 and the icons render fine in those environments. 
I don't know if it's XFCE or that I'm remotely accessing it, but I'd expect the icons to render either way.

**Screenshots**
![Image](https://github.com/user-attachments/assets/d6b16018-4225-4551-9e27-66ffab44fb7d)

**MediaElch Version:**
 - [ ] 2.12.1-dev (nightly)
 - [x] 2.12.0 (stable)

**Operating System:**
 - [ ] Windows
 - [ ] macOS
 - [x] Linux (distribution: Mint XFCE 22.1)

**Additional context**
I tried switching the theme from auto to dark, then light, and it didn't fix it.

bugwelle:
Hi,

thanks for reporting. Do you use our official PPA?

It looks as if SVG images can't be rendered properly. It could be that an SVG plugin is missing. I'll have to check.


Regards,
Andre

bugwelle:
Tested on Fedora by building MediaElch from scratch, and it seems to work just fine.

Could you maybe paste the command you used to install MediaElch? :)

bugwelle:
Sorry, I mixed up your issue with another one I was working on.

__I can reproduce your issue__.

Please install `libqt6svg6` to fix your issue:

```sh
sudo apt install libqt6svg6
```

It's missing as a runtime dependency for MediaElch, it seems.

To be fixed with the next debian packaging refactoring.

Regards,
Andre

dmackca:
Right on, @bugwelle ! That fixed it. I installed that dependency and now all of the icons are showing up as expected. Thank you

PR #1870, opened 2025-03-22
fix(debian: Add libqt6svg6 to the package dependencies

iluminat23:
It seems that ${shlibs:Depends} missed the dependency on libqt6svg6. Put it in the Depends to fix issues with svg images.

This should fix: https://github.com/Komet/MediaElch/issues/1859


bugwelle (approved):
Thanks. As far as I know, it was intended that dependencies are auto-discovered. For plugins such as SVG, however, that seems to not work properly. :)

Issue #1883, opened 2025-05-07
Dolby Vision not displayed


nkatsidis:
Latest Medialech Nightly Windows 10.

When a movie is Dolby Vision only HDR 10 displayed (in stream details).

can you please fix it so Dolby Vision is displayed.

nkatsidis:
if its dolby vision just add <hdrtype>dolbyvision</hdrtype>

thank you :)

Quppa:
I think this might be because the app is inspecting the value of `HDR_Format_Compatibility`, which for Dolby Vision is usually `HDR10` (the fallback for devices that don't support DV). https://github.com/bugwelle/MediaElch/blob/master/src/media/MediaInfoFile.cpp#L125

Looking at the output from `mediainfo` locally, the correct value (confusingly) seems to come from `HDR_Format_Commercial` for HDR10 and HDR10+ (which aren't distinct in Kodi's NFO format anyway) and `HDR_Format` for DV. I'll need to do some more testing over my media library to see how reliable this is. For the one HLG file I have handy, the `HDR_*` fields don't seem to be returned at all, but `transfer_characteristics` has `HLG / BT.2020 (10-bit)`.

That said, I don't see the field populated at all in the latest build - I can select a value manually and it will be saved in the .NFO, but it's left blank after hitting `Load stream details`.

bugwelle:
Thanks @Quppa 

The mediainfo fields are always a bit confusing. 😄 

How did you install MediaElch? What version of mediainfo do you use? :)

Quppa:
> The mediainfo fields are always a bit confusing. 😄

I went through my media library and checked the values for the various HDR fields. I think the logic should be something like:

```
if video["HDR_Format"].lower().contains("dolby vision") {
  return "dolbyvision";
} else if video["HDR_Format"].lower().contains("smpte st") {
  return "hdr10";
} else if video["transfer_characteristics"].lower().contains("hlg") {
  return "hlg";
}
```

In practice, the values I see are:
| Type | Mediainfo HDR_Format |
| ------ | ---------- |
| Dolby Vision (no HDR10 fallback) | `Dolby Vision` |
| Dolby Vision + HDR10 fallback | `Dolby Vision / SMPTE ST 2086` or `Dolby Vision / SMPTE ST 2086 / SMPTE ST 2086` |
| Dolby Vision + HDR10+ + HDR10 | `Dolby Vision / SMPTE ST 2086 / SMPTE ST 2094 App 4` |
[...]

PR #1899, opened 2025-07-07
fix: HDR10 detection for media files

bugwelle:
For #1883.


bugwelle:
> LGTM. I think the `HDR_Format_Compatibility` check is redundant but it doesn't really hurt, either.

Probably, yes. :)

PR #1917, opened 2025-10-31
fix: renamer applies to all subtitles

wwwehr:
use the subtitle filename extensions from settings so that idx and sub files get picked up by renamer

### DIFF
--- a/src/renamer/MovieRenamer.cpp
+++ b/src/renamer/MovieRenamer.cpp
@@ -215,6 +215,9 @@ MovieRenamer::RenameError MovieRenamer::renameMovie(Movie& movie)
                 }
 
                 // Handle all external subtitle files using preconfigured patterns
+                // FIXME: All of these subtitles should have been found by the movie-file-searcher.
+                // It shouldn't be necessary to add them in the renamer.
+                // This is a (hopefully) temporary workaround. See #1917
                 const mediaelch::FileFilter& subtitleFilters = Settings::instance()->advanced()->subtitleFilters();
                 QStringList subtitlePatterns;
                 for (const QString& pattern : subtitleFilters.fileGlob) {


src/file_search/movie/MovieDirectorySearcher.cpp
```diff
@@ -283,8 +284,27 @@ void MovieDiskLoader::createMovie(QStringList files)
         movie->controller()->loadData(Manager::instance()->mediaCenterInterface());
         if (discType == DiscType::Single) {
             QFileInfo mFi(files.first());
+
+            // Get subtitle filters from AdvancedSettings
+            const mediaelch::FileFilter& subtitleFilters = Settings::instance()->advanced()->subtitleFilters();
```
bugwelle: Just FYI: I try to avoid using the settings singleton via `Settings::instance()` because it makes testing more difficult (global state and all). I never got to test this class via our tests in `test/`, but this may make it more difficult to do so.

It's fine for now. However, I will probably refactor this coding a bit and pass the "subtitleFilters" in via constructor injection. :)

src/renamer/MovieRenamer.cpp
```diff
@@ -213,24 +214,97 @@ MovieRenamer::RenameError MovieRenamer::renameMovie(Movie& movie)
...
-                for (const QString &subFileName: currentDir.entryList(filters, QDir::Files |
-                QDir::NoDotAndDotDot)) { QString subSuffix = subFileName.mid(baseName.length()); QString newBaseName
-                = newFileName.left(newFileName.lastIndexOf(".")); QString newSubName = newBaseName + subSuffix;
-                    ui->results->append(tr("<b>Rename File</b> \"%1\" to \"%2\"").arg(subFileName).arg(newSubName));
-                    if (!dryRun) {
-                        if (!rename(currentDir.canonicalPath() + "/" + subFileName, currentDir.canonicalPath() + "/"
-                + newSubName)) ui->results->append("&nbsp;&nbsp;<span style=\"color:#ff0000;\"><b>" + tr("Failed") +
-                "</b></span>"); else FilmFiles.append(newSubName);
+                // Handle all external subtitle files using preconfigured patterns
+                const mediaelch::FileFilter& subtitleFilters = Settings::instance()->advanced()->subtitleFilters();
```
bugwelle: hmmm...

I do wonder why we need to look for new subtitles here. Shouldn't they already be part of `movie.subtitles()`? After all, you've adapted `src/file_search/movie/MovieDirectorySearcher.cpp`.

wwwehr: it doesn't seem to find them as "movie.subtitles()" and so I think my new approach just deals with them as an aside. when I dump the vector right before line 218 `movie.subtitles()` shows empty. wdyt? I care less about the proper tracking of the movie assets than I do the renamer "just working"

src/renamer/MovieRenamer.cpp
```diff
@@ -213,24 +214,97 @@ MovieRenamer::RenameError MovieRenamer::renameMovie(Movie& movie)
...
-                    filters << baseName + extra;
-                for (const QString &subFileName: currentDir.entryList(filters, QDir::Files |
-                QDir::NoDotAndDotDot)) { QString subSuffix = subFileName.mid(baseName.length()); QString newBaseName
-                = newFileName.left(newFileName.lastIndexOf(".")); QString newSubName = newBaseName + subSuffix;
-                    ui->results->append(tr("<b>Rename File</b> \"%1\" to \"%2\"").arg(subFileName).arg(newSubName));
-                    if (!dryRun) {
-                        if (!rename(currentDir.canonicalPath() + "/" + subFileName, currentDir.canonicalPath() + "/"
-                + newSubName)) ui->results->append("&nbsp;&nbsp;<span style=\"color:#ff0000;\"><b>" + tr("Failed") +
-                "</b></span>"); else FilmFiles.append(newSubName);
+                // Handle all external subtitle files using preconfigured patterns
```
bugwelle:
```suggestion
                // Handle all external subtitle files using preconfigured patterns
                // FIXME: All of these subtitles should have been found by the movie-file-searcher.
                // It shouldn't be necessary to add them in the renamer.
                // This is a (hopefully) temporary workaround. See #1917
```

bugwelle (approved):
LGTM; please add the one suggestion, then I'll merge this. :)

Issue #1920, opened 2025-11-15
Imdb scraper freezes search screen, 'scraper did not respond with any data'


Tachikomatic-C:
Any time I now try and load information for a movie, it gets stuck on search. the spinning wheel keeps going round and round and it is stuck on IMDb. I cannot click on anything or change any settings i.e. to go back to TheMovieDB

**Expected behavior**
Expect to be able to change the scraper/click on a dropdown

**Screenshots**

<img width="911" height="945" alt="Image" src="https://github.com/user-attachments/assets/50a08bfa-1b6c-4d60-8efc-18a567e129a8" />

**MediaElch Version:**

 - [x] 2.12.0 (stable)

**Operating System:**
 - [x] Windows 11

**Additional context**

ticao2:
IMDB is having problems. They changed the entire WEB interface.
See here https://github.com/Komet/MediaElch/issues/1889

ACR-Jeff:
> Where is this located in windows?

Been a while since I messed with Windows
I am going to guess in  `C:/Programs Files/MediaElch` directory
or possibly in your profile directory aka user home folder and within the hidden directory `appdata`

Be sure to always backup files before modifying or removing

PR #1954, opened 2026-03-17
fix(ui): keep scraper combobox enabled during movie search

### LINKED ISSUE #1945: Movie Scraper other than IMDB no longer selectable
**Describe the bug**
When trying to get info for a movie by selecting "load Information" in the context menu, the IMDB scraper shows an error message like "Scraper sent no data" and the Combo bow to select another scraper like TMDB remains disabled.
**To Reproduce**
1. Right Mouse Button on Movie title
2. from context menu, select "load Information"

**Expected behavior**
When issues with a scraper occur, there should be a non-response timeout so that funtions do not remain disabled

**Screenshots**
-

**MediaElch Version:**
 - [ ] 2.12.1-dev (nightly)
 - [x] 2.12.0 (stable)

**Operating System:**
 - [ ] Windows
 - [x] macOS

**Additional context**
-

Arny80Hexa:
## Problem

When a movie search is in progress, the scraper combobox is disabled. If the active scraper fails (e.g. IMDB returning no data due to changed markup), users cannot switch to a working scraper (e.g. TMDb) until the network request times out (~10 seconds). This makes the search dialog appear frozen and unresponsive.

Additionally, since `doKill()` is not overridden in scraper search jobs, calling `kill()` on a running search job silently fails — the old job continues running in the background. When it eventually completes, its results (or error message) can overwrite the results of a newer search.

Fixes #1945
Relates to #1920

## Changes

**`src/ui/movies/MovieSearchWidget.cpp`**

1. **Keep scraper combobox enabled during search** — Only disable it during custom scraper multi-step workflows (where `isCustomScrapingInProgress()` is true). For normal searches, the combobox stays enabled so users can switch scrapers at any time. Switching mid-search is safe because `startSearch()` already calls `abortAndClearResults()` before starting a new search.

2. **Discard stale search results** — Before processing results in `onShowResults()`, check whether the search job is still the current one. If a newer search has been started, discard the stale results silently. This prevents old error messages from overwriting current search results.

## Testing

Tested manually against v2.12.0 behavior:
- IMDB scraper selected → search returns "0 results" (scraper is broken)
- **Before fix:** Scraper dropdown disabled during search, UI appears frozen
- **After fix:** Dropdown stays clickable, user can immediately switch to TMDb

---
*Developed with AI assistance (Claude Code).*

### DIFF
--- a/src/ui/movies/MovieSearchWidget.cpp
+++ b/src/ui/movies/MovieSearchWidget.cpp
@@ -227,7 +227,6 @@ void MovieSearchWidget::onShowResults(mediaelch::scraper::MovieSearchJob* search
         // A newer search has been started; discard these stale results.
         return;
     }
-    m_currentSearchJob = nullptr;
 
     if (searchJob->wasKilled()) {
         // If it was killed, don't report anything, but reset the search bar.
--- a/src/ui/movies/MovieSearchWidget.cpp
+++ b/src/ui/movies/MovieSearchWidget.cpp
@@ -96,7 +96,10 @@ void MovieSearchWidget::startSearch()
     emit sigMovieSelectionChanged(false);
 
     abortAndClearResults();
-    ui->comboScraper->setEnabled(false);
+    // Only disable the scraper combobox during custom scraper workflows.
+    // For normal searches, keep it enabled so users can switch scrapers
+    // even while a search is in progress (e.g. if the current scraper is broken).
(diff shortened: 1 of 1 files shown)


bugwelle (changes requested):

bugwelle (approved):
Tested locally, LGMT.

PR #1912, opened 2025-10-26
fix(imdb): Fix IMDB scraping

bugwelle:
IMDB's layout changed drastically, meaning we can no longer use RegEx "parsing" on their HTML.  Instead, use their JSON data instead.

Fix #1889

TODO:
- [x] movies
  - [x] outline
  - [x] overview
  - [x] title
  - [x] original title
  - [x] actors
  - [x] certification
  - [x] genres
  - [x] directors
  - [x] writers
  - [x] runtime
  - [x] rating
  - [x] release date
  - [x] tags
  - [x] studios
  - [x] countries
  - [x] trailer (new!)
  - [x] poster
  - [ ] …
- [x] TV shows
  - [x] search
  - [x] scrape data

PR #1955, opened 2026-03-17
refactor(imdb): parse search results from JSON instead of HTML regex

Arny80Hexa:
## Summary

Replace the regex-based HTML parser in `ImdbSearchPage::parseSearch()` with a JSON parser that reads from IMDB's embedded `__NEXT_DATA__` script tag.

The current regex parser still works, but it is fragile — it depends on IMDB's HTML markup staying compatible. IMDB has already migrated their frontend to Next.js/React, and the HTML structure has changed before (breaking the parser in earlier versions). The `__NEXT_DATA__` JSON is the structured data source that IMDB's own frontend renders from, making it a more stable parsing target.

This is the same approach already used by `ImdbJsonParser` for title detail pages (introduced in PR #1912).

Relates to #1920

## Changes

**`src/scrapers/imdb/ImdbSearchPage.cpp`**

Replace the regex-based HTML extraction with JSON parsing of the `__NEXT_DATA__` script tag.

JSON path: `props.pageProps.searchResults.titleResults.titleListItems[]`

Each item provides `titleId`, `titleText`, and `releaseYear` — exactly the fields needed for search results.

## Testing

Tested manually:
- **Search by IMDB ID** (e.g. `tt1254207`): finds the movie, shows preview with poster and description
- **Search by title** (e.g. `Inception`): returns results with correct titles and years
- **TV show search**: uses the same parser, also works
- **Empty/invalid queries**: returns 0 results gracefully

---
*Developed with AI assistance (Claude Code / Opus 4.6).*

### DIFF
--- a/src/scrapers/imdb/ImdbSearchPage.cpp
+++ b/src/scrapers/imdb/ImdbSearchPage.cpp
@@ -1,38 +1,71 @@
 #include "ImdbSearchPage.h"
 
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
 #include <QRegularExpression>
 
-#include "scrapers/ScraperUtils.h"
+#include "log/Log.h"
 
 namespace mediaelch {
 namespace scraper {
 
 QVector<ImdbSearchPage::SearchResult> ImdbSearchPage::parseSearch(const QString& html)
 {
-    // Search result table from "https://www.imdb.com/search/title/?title=..."
-    // The results may contain the user's locale, e.g. `/de/title/…`.
-    static const QRegularExpression rx(R"(<a href="/(?:\w{2,4}/)?title/(tt[\d]+)/[^>]+>(.+)</a>.*(\d{4})[–<])",
-        QRegularExpression::DotMatchesEverythingOption | QRegularExpression::InvertedGreedinessOption);
-    // Entries are numbered: Remove Number.
-    static const QRegularExpression listNo(
-        R"(^\d+\.\s+)", QRegularExpression::DotMatchesEverythingOption | QRegularExpression::InvertedGreedinessOption);
-
     QVector<SearchResult> results;
-    QRegularExpressionMatchIterator matches = rx.globalMatch(html);
-    QRegularExpressionMatch match;
-    while (matches.hasNext()) {
-        match = matches.next();
-        if (match.hasMatch()) {
-            QString title = normalizeFromHtml(match.captured(2));
-            title.remove(listNo);
-            SearchResult result;
-            result.title = title;
-            result.identifier = match.captured(1);
-            result.released = QDate::fromString(match.captured(3), "yyyy");
-            results.push_back(std::move(result));
-        }
+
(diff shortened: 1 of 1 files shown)


Arny80Hexa:
Thanks for the hint! I checked the master branch and it still uses the regex-based HTML parser. The nightly PPA on Launchpad also appears to be inactive. Could you point me to the build you're referring to? Happy to compare approaches if there's already a fix somewhere.

kolbdog323:
> Thanks for the hint! I checked the master branch and it still uses the regex-based HTML parser. The nightly PPA on Launchpad also appears to be inactive. Could you point me to the build you're referring to? Happy to compare approaches if there's already a fix somewhere.

Download the nightly build from the official media elch page heres the link https://mediaelch.github.io/mediaelch-doc/download.html

Arny80Hexa:
Thanks for the link! I checked all available snapshot builds (July 2025 through February 2026) — they're all built from master and all contain the old regex-based HTML parser in `ImdbSearchPage.cpp`. None of them include a JSON-based search parser.

Could you verify that IMDB search actually returns results in the snapshot you're using? The issue is specifically with the search result parsing, not the title detail page (which was already fixed in PR #1912).

kolbdog323:
> Thanks for the link! I checked all available snapshot builds (July 2025 through February 2026) — they're all built from master and all contain the old regex-based HTML parser in `ImdbSearchPage.cpp`. None of them include a JSON-based search parser.
> 
> Could you verify that IMDB search actually returns results in the snapshot you're using? The issue is specifically with the search result parsing, not the title detail page (which was already fixed in PR #1912).

Sure here you go.
Screenshot from 2026-03-18 11-35-20
Screenshot from 2026-03-18 11-36-37
Screenshot from 2026-03-18 11-37-37
Screenshot from 2026-03-18 11-37-54

kolbdog323:
Screenshot from 2026-03-18 11-41-09

kolbdog323:
Did that help any I used the newest nightly build from February 20th

Arny80Hexa:
You're right — I downloaded and tested the Feb 20th snapshot myself, and IMDB title search does work with the current regex parser. Thanks for pushing back on this, it helped us verify properly.

That said, the regex approach is fragile — it depends on IMDB's HTML structure staying compatible, which has broken before. The JSON parser reads from the structured `__NEXT_DATA__` that IMDB embeds on every page, which is more stable since it's the data source IMDB's own frontend uses. I'll update the PR description to reflect that this is a robustness improvement rather than a critical fix.

Arny80Hexa:
Closing this PR — IMDB has rolled out AWS WAF JavaScript challenges on all HTML endpoints (`x-amzn-waf-action: challenge`), returning HTTP 202 with empty body. This affects the search page, title page, and reference page. Since this PR parses `__NEXT_DATA__` JSON from the HTML response, it no longer works when the HTML itself is blocked.

I've opened a new issue to track the problem and discuss a path forward using IMDB's Suggest API (search) and GraphQL API (details), which both still work reliably. See #1966.

PR #1959, opened 2026-03-17
feat(custom-scraper): auto-resolve IMDB ID from TMDb in multi-step search

Arny80Hexa:
## Problem

When the Custom Movie Scraper uses TMDb as the title scraper and IMDB as a secondary scraper, the multi-step search forces users to manually re-search on IMDB with the English title. This is particularly frustrating for non-English users who search with localized titles (e.g. "Die Verurteilten" instead of "The Shawshank Redemption").

## Fix

Before switching to the IMDB search step, resolve the IMDB ID from the TMDb API using the already-known TMDb movie ID. The IMDB search then uses this ID directly, finding the movie in a single step without manual intervention.

**`src/ui/movies/MovieSearchWidget.h` / `MovieSearchWidget.cpp`**
- Add `resolveImdbIdFromTmdb()` method that queries TMDb's `/movie/{id}` endpoint to get the `imdb_id`
- In `onResultDoubleClicked()`, when the next scraper is IMDB and the title scraper was TMDb, resolve the IMDB ID before switching
- Set the resolved IMDB ID as the search text so IMDB can do a direct ID lookup

## Testing

Custom Scraper configured with Title=TMDb, other fields=IMDb:

| Step | Before | After |
|------|--------|-------|
| 1. Search "Die Verurteilten 1994" | TMDb finds movie | TMDb finds movie (unchanged) |
| 2. Switch to IMDB | Search field: "Die Verurteilten 1994" → 0 results | Search field: "tt0111161" → 1 result found automatically |
| 3. User action needed | Must manually type English title | None — proceeds automatically |

The IMDB ID resolution adds ~1-2 seconds to the scraper switch (single TMDb API call), which is negligible compared to the manual search it replaces.

---
*Developed with AI assistance (Claude Code / Opus 4.6).*

### DIFF
--- a/src/ui/movies/MovieSearchWidget.cpp
+++ b/src/ui/movies/MovieSearchWidget.cpp
@@ -11,14 +11,7 @@
 #include "ui/movies/MoviePreviewAdapter.h"
 #include "utils/Meta.h"
 
-#include "scrapers/tmdb/TmdbApi.h"
-
 #include "log/Log.h"
-
-#include <QJsonDocument>
-#include <QJsonObject>
-#include <QNetworkReply>
-#include <QNetworkRequest>
 
 MovieSearchWidget::MovieSearchWidget(QWidget* parent) : QWidget(parent), ui(new Ui::MovieSearchWidget)
 {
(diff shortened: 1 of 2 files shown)


bugwelle (commented):
The idea is good and I'll merge it if one small remark can be fixed.

Even though I wanted to decouple the search widget from the actual scrapers, it seems that for the best user interaction, that is not necessarily the best way. I'm fine with depending on e.g. the TMDb API here, so that we can avoid the direct network request.

bugwelle (changes requested):
see previous comment

Arny80Hexa:
Thanks for the feedback! I agree we should reuse the TMDb API classes instead of the custom network request.

I see three possible approaches and would appreciate your guidance on which you prefer:

1. **Add a public accessor** to `TmdbMovie` for its `TmdbApi` instance (smallest change, but exposes internals)
2. **Create a standalone `TmdbApi` instance** in the widget (works but redundant)
3. **Use a `MovieScrapeJob`** via the existing `TmdbMovie` scraper from `CustomMovieScraper::titleScraper()` to resolve the IMDB ID through the established code path (cleanest reuse, but more refactoring)

I'm leaning towards option 3 as it fully reuses the existing infrastructure. Let me know what you think and I'll update the PR accordingly.

bugwelle (approved):
LGTM. Thanks!

PR #1974, opened 2026-03-20
fix(videobuster): strip year from search query

### LINKED ISSUE #1973: VideoBuster search fails when query contains a year
VideoBuster search returns 0 results when the search query contains a year (e.g. "Inception 2010"), because the year is passed as part of the search string to the VideoBuster website, which does not support it.

`MovieSearchJob::extractTitleAndYear()` exists to strip the year from search queries, but has two problems:
1. It is not used by `VideoBusterSearchJob`
2. The method itself has a bug — it returns the full match string instead of the captured title/year groups

**Affected:** VideoBuster (confirmed). TMDb, IMDB and OMDb handle years internally and are not affected.

Related: #1971

Analyzed with AI assistance (Claude Code / Opus 4.6).

Arny80Hexa:
Fix VideoBuster search returning 0 results when the query contains a year (e.g. "Inception 2010"). VideoBuster does not support year in the search string.

Also fixes a bug in `MovieSearchJob::extractTitleAndYear()` where `captured(0)` (full match) was returned instead of `captured(1)` (title) and `captured(2)` (year).

Fixes #1973

Developed with AI assistance (Claude Code / Opus 4.6).

### DIFF
--- a/src/scrapers/movie/MovieSearchJob.cpp
+++ b/src/scrapers/movie/MovieSearchJob.cpp
@@ -18,9 +18,9 @@ QPair<QString, QString> MovieSearchJob::extractTitleAndYear(const QString& query
         rxYear.setPatternOptions(QRegularExpression::InvertedGreedinessOption);
         auto match = rxYear.match(query);
         if (match.hasMatch()) {
-            QString searchTitle = match.captured(0);
-            QString searchYear = match.captured(1);
-            return {searchTitle, searchTitle};
+            QString searchTitle = match.captured(1);
+            QString searchYear = match.captured(2);
+            return {searchTitle, searchYear};
         }
     }
     return {};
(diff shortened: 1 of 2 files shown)


src/scrapers/movie/MovieSearchJob.cpp
```diff
@@ -18,9 +18,9 @@ QPair<QString, QString> MovieSearchJob::extractTitleAndYear(const QString& query
         rxYear.setPatternOptions(QRegularExpression::InvertedGreedinessOption);
         auto match = rxYear.match(query);
         if (match.hasMatch()) {
-            QString searchTitle = match.captured(0);
-            QString searchYear = match.captured(1);
-            return {searchTitle, searchTitle};
+            QString searchTitle = match.captured(1);
```
bugwelle, line 21: Indeed, this was just wrong before. Wow. Copy&Paste at its best. Thanks!

bugwelle (approved):

PR #1976, opened 2026-03-20
feat(custom): add separate Outline field to Custom Movie Scraper

### LINKED ISSUE #1975: Custom Movie Scraper: add separate outline/short summary field
The Custom Movie Scraper has no way to assign a separate scraper for the movie outline (short summary). Currently only "Overview" (plot) can be assigned to a scraper, and the outline field either stays empty or gets a copy of the plot via the "Use plot for outline" fallback setting.

However, the internal data model already supports separate fields — `Movie::outline()` and `Movie::overview()` are distinct, and the Kodi NFO writer correctly outputs both as `<outline>` and `<plot>`. The missing piece is a "Zusammenfassung" / "Outline" row in the Custom Movie Scraper configuration table, allowing users to pick a different scraper for the short summary.

**Use case:** German users could set Plot → VideoBuster (detailed German descriptions) and Outline → TMDb (shorter German summary), getting two meaningfully different texts.

Analyzed with AI assistance (Claude Code / Opus 4.6).

Arny80Hexa:
Allow users to assign a different scraper for the movie outline (short summary) than for the plot. The Kodi NFO writer already outputs both as separate `<outline>` and `<plot>` tags — this change exposes that capability in the Custom Movie Scraper settings.

When no dedicated Outline scraper is configured, the existing "use plot for outline" fallback setting continues to work as before.

Fixes #1975

Developed with AI assistance (Claude Code / Opus 4.6).


src/data/movie/Movie.cpp
```diff
@@ -82,7 +82,8 @@ void Movie::clear()
           << MovieScraperInfo::Banner        //
           << MovieScraperInfo::Thumb         //
           << MovieScraperInfo::ClearArt      //
-          << MovieScraperInfo::TvShowLinks;
+          << MovieScraperInfo::TvShowLinks
+          << MovieScraperInfo::Outline;
```
bugwelle:
```suggestion
          << MovieScraperInfo::TvShowLinks   //
          << MovieScraperInfo::Outline;
```

FYI: Please run `./scripts/run_clang_format.sh` and/or `./scripts/quick_checks.sh` once.  The trailing comment is a "hack" to make clang-format auto-format these lines.

src/scrapers/movie/MovieMerger.cpp
```diff
@@ -64,9 +64,12 @@ void copyDetailToMovie(Movie& target,
     }
     case MovieScraperInfo::Overview: {
         target.setOverview(source.overview());
+        break;
+    }
+    case MovieScraperInfo::Outline: {
         if (!source.outline().isEmpty()) {
             target.setOutline(source.outline());
-        } else if (usePlotForOutline) {
+        } else if (!source.overview().isEmpty()) {
```
bugwelle: Why remove the `usePlotForOutline` check?


```suggestion
        } else if (target.outline().isEmpty() && usePlotForOutline && !source.overview().isEmpty()) {
```

test/unit/scrapers/testMovieMerger.cpp
bugwelle, line 29: Could you copy this test case and test:

- no outline requested
- "usePlotForOutline" being true

=> ensure that outline isn't set

src/scrapers/movie/MovieMerger.cpp
```diff
@@ -198,8 +201,18 @@ void copyDetailsToMovie(Movie& target,
...
+            continue; // handled below
+        }
         copyDetailToMovie(target, source, detail, usePlotForOutline, ignoreDuplicateOriginalTitle);
     }
+
+    if (details.contains(MovieScraperInfo::Outline)) {
+        copyDetailToMovie(target, source, MovieScraperInfo::Outline, usePlotForOutline, ignoreDuplicateOriginalTitle);
+    } else if (usePlotForOutline && target.outline().isEmpty() && !target.overview().isEmpty()) {
+        target.setOutline(target.overview());
+    }
```
bugwelle: I'd prefer this logic to be inside `copyDetailToMovie` instead of here. `usePlotForOutline` is rather a `usePlotForOutlineIfOutlineIsEmpty` :D

bugwelle (changes requested):
Cool! Definitely a good addition. I have a few remarks regarding the movie merger. Would be great if you could incorporate that. :)

Arny80Hexa:
Addressed all review comments:
- `Movie.cpp`: added trailing `//` for clang-format alignment
- `MovieMerger.cpp`: restored `usePlotForOutline` check with `target.outline().isEmpty()` guard, moved all logic into `copyDetailToMovie` (removed special handling from the caller)
- Added test case: outline not requested + `usePlotForOutline=true` → outline stays empty

Also rebased on current master. Ran `run_clang_format.sh` and `quick_checks.sh` — will make sure to run both before every future push.

Side note: we noticed that `run_clang_format.sh` expects version 19, while current LLVM is at 22. We installed 19 locally, so no issue on our end. If upgrading to a newer clang-format version is something you'd like to do at some point, we'd be happy to take on that reformatting commit.

bugwelle:
> Side note: we noticed that run_clang_format.sh expects version 19, while current LLVM is at 22. We installed 19 locally, so no issue on our end. If upgrading to a newer clang-format version is something you'd like to do at some point, we'd be happy to take on that reformatting commit.

Yeah. I match my local system :D 
But I'll have to update it soon. The version check exists because with each clang-format version, the style changes.

bugwelle (approved):
Awesome, thank you very much! :)

Issue #1992, opened 2026-05-15
network error


DrRamesh:
Title: Built-in TMDb API key returns HTTP 400 - key appears revoked

MediaElch version: 2.12.1-dev (2026-04-27)
OS: Windows 11 Pro 25H2

The built-in TMDb API key (5d832bdf69dcb884922381ab01548d5b) 
returns HTTP 400 from TMDb's API. The key appears to have been 
revoked. There is no UI option to enter a personal TMDb API key.

Please either update the built-in key or add a settings field 
to allow users to enter their own TMDb API key.

ghost:
Have a network error too:

_Network Error: Could not load the requested resource._

kolbdog323:
Did you gusy try the nightly version it works fine for me for TMDB api i cna fetch all the info

AnselmD:
Same network error:

MediaElch 2.12.1-dev - Benzar ( MediaElch_win_Qt6_2.12.1-dev_2026-04-27_00-40_git-master-c4d47773)

Compiled with Qt version 6.11.0 (64 Release, GCC 11.5.0)
Using Qt version 6.11.0
System: Windows 10 Version 22H2 (x86_64-little_endian-llp64)

and

MediaElch 2.12.0 - Benzar

thefrip:
Same here but doesn't seem an issue with the key but with the parameters of the call which includes twice the "include_adult" parameter. Maybe the updated TMDB API is more sensitive and doesn't accept to receive twice the same parameter (I tested by hand and it works when not repeating the parameter).

ghost:
@thefrip 
Do you have an easy workaround for that to get it fixed in the meantime?

AnselmD:
> Same here but doesn't seem an issue with the key but with the parameters of the call which includes twice the "include_adult" parameter. Maybe the updated TMDB API is more sensitive and doesn't accept to receive twice the same parameter (I tested by hand and it works when not repeating the parameter).

Yes, this is the problem, curl with double include_adult does not work ( i removed the api_key):

C:\WINDOWS\system32>curl.exe -I "https://api.themoviedb.org/3/search/movie?page=1&query=testFilm&include_adult=false&include_adult=false&api_key=xxxxxx&language=en-US&region=US"
HTTP/1.1 400 Bad Request
Content-Type: application/json;charset=utf-8
Content-Length: 111
Connection: keep-alive
Date: Fri, 15 May 2026 18:22:31 GMT
Server: openresty
[...]

PR #1995, opened 2026-05-20
[TMDb] Fix duplicate include_adult parameter in movie search

RawScape:
Fixes Issue #1992.

The `include_adult` query parameter was being appended twice to TMDb
movie search URLs, causing TMDb to reject the requests with HTTP 400.

In `TmdbMovieSearchJob::doStart()`, `includeAdult` was passed both as
a direct argument to `getMovieSearchUrl()` (which adds it to the
query) and again via the `UrlParameterMap` using
`ApiUrlParameter::INCLUDE_ADULT`. Removed the redundant map entry in
both code paths so the parameter is now sent exactly once.

Testing: Verified the final request URL contains `include_adult`
exactly once and that movie search returns results again instead of
HTTP 400.

### DIFF
--- a/src/scrapers/movie/tmdb/TmdbMovieSearchJob.cpp
+++ b/src/scrapers/movie/tmdb/TmdbMovieSearchJob.cpp
@@ -46,10 +46,7 @@ void TmdbMovieSearchJob::doStart()
         url.swap(newUrl);
 
     } else {
-        QUrl newUrl(m_api.getMovieSearchUrl(searchStr,
-            config().locale,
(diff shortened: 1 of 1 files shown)


snostormjp:
I built this on Arch Linux for AUR using the file from this PR and it fixes the issues with not being able to scrape movies.

bugwelle (approved):

bugwelle:
Thanks for the contribution! Sorry for the delay. I will update https://github.com/Komet/MediaElch/issues/1710 soon.

PR #1999, opened 2026-06-18
refactor: constructor-inject persistence classes + better testing

bugwelle:
This commit makes the persistence classes virtual, so that we can
better test them in the future.

Furthermore, it constructor-injects the persistence classes where
needed.


bugwelle:
- [x] TODO: segfault somewhere

PR #2000, opened 2026-06-18
refactor: Use Key/Value for all settings, i.e. common interface

bugwelle:
I want to streamline our settings handling: Use the common key/value pair mechanism.


bugwelle:
tested locally; to be merged in the next few days

PR #2006, opened 2026-07-21
Added Abort Scrape Function

purpleperkin:
While trying to scrape IMDB, i got caught in a loop, where no other scraper was able to be selected. added an abort function.


purpleperkin:
IMDB APK issue still exists, but at least you can abort the task and switch to a different scraper

PR #2007, opened 2026-08-02
i18n: Add Russian translation for desktop entry and AppStream data

wehrwolfmann:
The launcher entry and the AppStream data are about the only user-visible strings MediaElch ships that are not translated into anything — they are not part of the Transifex resource, so they can only arrive as a patch. On a Russian desktop that shows: the application itself speaks Russian, while the menu and the software centre still say "Media Manager".

I translated `GenericName` instead of adding a `Comment`. Plasma's launcher and KRunner take the generic name first and only fall back to `Comment` when it is empty, so a `Comment[ru]` on its own would have changed nothing — I tried it locally and the menu still read "Media Manager"; with `GenericName[ru]` it reads "Менеджер медиатеки". That also keeps the entry free of a second field saying the same thing twice.

The wording follows the Russian UI translation that is already on Transifex (фильмы, сериалы, концерты, музыка, файлы NFO), so the menu and the application use the same terms. `desktop-file-validate` stays clean and `appstreamcli validate` reports exactly the same notes as before the change; `appstreamcli compose` picks up all three translated strings into the catalogue.

### DIFF
--- a/data/desktop/MediaElch.desktop
+++ b/data/desktop/MediaElch.desktop
@@ -6,3 +6,4 @@ Terminal=false
 Exec=MediaElch
 Categories=AudioVideo;AudioVideoEditing;Database;Qt;
 GenericName=Media Manager
+GenericName[ru]=Менеджер медиатеки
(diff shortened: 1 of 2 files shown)


wehrwolfmann:
Gentle nudge on this one — it is two lines in the desktop entry and two in the metainfo, no code involved. Rebased cleanly on main as of today. Happy to adjust the wording if you would rather have something else there.

bugwelle (approved):
Thank you very much for this PR! For reasons why it took so long, please see https://github.com/Komet/MediaElch/issues/1710 including the latest update. :)

